PR #1, opened 2016-04-08
Standardized whitespace, removed vestigial comments

EFox2413:
Adjusted braces to follow this syntax:

```javascript
function myFunction() {
    ...
}
```

Adjusted 1 function so variables did not flow past 100 column width
```javascript
function myFunction(var1, var2, var3, var4, var5,
    var6, var7, var8, var9) {
    ...
}
```

deleted (seemingly) vestigial comments of outdated functions and misplaced documentation (1 instance)

deleted unnecessary white-space between functions (made it so that most times there is but 1 line of white-space in-between functions and some comments)

Deleted trailing white-space

Automated standard indentation linter

### DIFF
--- a/Initium-Odp/war/odp/javascript/script.js
+++ b/Initium-Odp/war/odp/javascript/script.js
@@ -1,60 +1,58 @@
 window.popupsNum = 0;
 window.popupsOpen = 0;
 window.popupsArray = new Array();
-
 window.singlePostFormSubmitted = false;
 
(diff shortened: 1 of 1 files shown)


Emperorlou:
I don't think this is a good idea. It breaks diffs, it creates massive conflicts with existing code changes, and most of the time the whitespace is actually done on purpose to make it easy to find major sections of the file when paging through quickly.

PR #5, opened 2016-04-13
SetLeader Bugfixes and SetLabel implementation

NikeJoshua:


Initium-Odp/src/com/universeprojects/miniup/server/commands/CommandSetLabel.java
```diff
@@ -0,0 +1,91 @@
...
+	}
+
+	@Override
+	public void run(Map<String, String> parameters) throws UserErrorMessage 
+	{
+		// Get the DB access object. We use this for most DB communication. It is a place for reusable methods that access the DB.
+		ODPDBAccess db = getDB();
+
+		// Check if we're logged in or not. We shouldn't be able to see this since commands cannot be used unless the user is logged in.
+		if (db.isLoggedIn(request)==false)
```
Emperorlou: I THINK the isLoggedIn() call is not necessary because the Command class always checks if logged in and throws if not. Please test to make sure this assumption is correct before taking any action tho.

Initium-Odp/src/com/universeprojects/miniup/server/commands/CommandSetLabel.java
```diff
@@ -0,0 +1,91 @@
...
+		Long maxWeight = (Long)item.getProperty("maxWeight");
+		if (maxWeight==null || maxWeight<=0)
+			throw new UserErrorMessage("This is not a storage item and thus you cannot relabel it.");
+
+		// Verify that the item is either in the caller's inventory, or at the same location.
+		Key containerKey = (Key)item.getProperty("containerKey");
+		if (containerKey==null)
+			throw new UserErrorMessage("You can only relabel items in your immediate vicinity.");
+		CachedEntity character = db.getCurrentCharacter(request);
+		if (containerKey.equals(character.getKey())==false && containerKey.equals((Key)character.getProperty("locationKey"))==false)
```
Emperorlou: An unfortunate quirk with using remote API is that keys don't compare properly. Please compare the Kind and the ID instead. I believe we have a method in GameUtils that does a proper equals comparison for keys so you don't have to write it out each time.

Yep just checked. GameUtils.equals()

Initium-Odp/src/com/universeprojects/miniup/server/commands/CommandSetLabel.java
```diff
@@ -0,0 +1,91 @@
...
+			throw new UserErrorMessage("You can only relabel items in your immediate vicinity.");
+		
+		// Assign new label.
+		try {
+			String label = parameters.get("label");
+			if (label==null || label.trim().equals(""))
+				item.setProperty("label", null);
+			else
+				item.setProperty("label", WebUtils.htmlSafe(label.trim()));
```
Emperorlou: Better to store the raw data in the database and let the jsps worry about what is html safe.

Initium-Odp/src/com/universeprojects/miniup/server/commands/CommandSetLabel.java
NikeJoshua: Just tested with a manually created command URL when logged out, and indeed gets thrown prior to this check, so will remove these for next update. The output is a regular errorMessage json, but since people aren't logged in, there's no clean handler for this and it just gets displayed as such:
{"errorMessage":"You are not currently logged in. \u003ca href\u003d\u0027login.jsp\u0027\u003eClick here\u003c\/a\u003e to log in before doing anything else."}
Maybe better to just write direct html output here?

Initium-Odp/src/com/universeprojects/miniup/server/commands/CommandSetLeader.java
```diff
@@ -45,7 +45,7 @@ public void run(Map<String, String> parameters) throws UserErrorMessage
 		Long charId;
 		try {
 			charId = Long.parseLong(parameters.get("charId"));
-		} catch (Exception e) {
```
Emperorlou: I would catch any exception here like you had before since any exception would always mean a malformed charId anyway.

Initium-Odp/src/com/universeprojects/miniup/server/commands/CommandSetLabel.java
Emperorlou: Ok don't worry about it then. I'll see what I can do when I have a chance.
Probably just need to update the javascript a bit.

PR #6, opened 2016-04-14
reloadPagePopup fix

NikeJoshua:


Emperorlou:
I don't think this was the problem. Here is the old reload function before iframe support was added:

```
function reloadPagePopup(quietly)
{
	if (currentPopupStackIndex==0)
		return;
	
	var pagePopupId = "page-popup"+currentPopupStackIndex;
	
	var url = $("#"+pagePopupId+"-content").attr("src");
	if (quietly)
	{}
	else
		$("#"+pagePopupId).html("<div id='"+pagePopupId+"-content' src='"+url+"' class='page-popup'><img id='banner-loading-icon' src='javascript/images/wait.gif' border=0/></div><div class='page-popup-glass'></div><a class='page-popup-Reload' onclick='reloadPagePopup()' style='font-family:Lucida Sans'>&#8635;</a><a class='page-popup-X' onclick='closePagePopup()'>X</a>");
	$("#"+pagePopupId+"-content").load(url);
}
```

PR #9, opened 2016-04-14
Some script.js fixes

NikeJoshua:
- Small typo in create campsite message
 - Several shortcut fixes
 - Reduced nonsensical popup stacking
 - Some minor code cleanup

### DIFF
--- a/Initium-Odp/war/odp/javascript/script.js
+++ b/Initium-Odp/war/odp/javascript/script.js
@@ -465,7 +465,7 @@ function createCampsite()
 		if (name!=null && name!="")
 		{
 			window.location.href='ServletCharacterControl?type=createCampsite&name='+encodeURIComponent(name);
-			popupPermanentOverlay("Creating a new campsite..", "You are hard at work setting up a new camp. Make sure you defend it or it wont last long!");
+			popupPermanentOverlay("Creating a new campsite..", "You are hard at work setting up a new camp. Make sure you defend it or it won't last long!");
 			localStorage.setItem("campsiteName", name);
 		}
 	});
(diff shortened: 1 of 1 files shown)


Initium-Odp/war/odp/javascript/script.js
```diff
@@ -1481,23 +1483,28 @@ $(document).keyup(function(event){
 		var buttonToPress = $('[shortcut="'+event.which+'"]');
 		if (buttonToPress.length>0)
 		{
-			window.disableShortcuts = true;
+			if (buttonToPress[0].hasClass("main-button-half")==false)
+				window.disableShortcuts = true;
```
Emperorlou: Why this?

NikeJoshua: It keeps the shortcuts available if a key for one of the Nearby XXX popups was used. Previously, opening one of these popups and then closing it left you stuck without shortcuts. I also thought this was a nicer way to do it than reenabling on close, as this way you can still use the shortcuts while the popup is open (Ex: Done reordering your house inventory? Just press 1 and you're back in Aera!)

Emperorlou:
Sorry I think Anton fixed the iframe issue as well creating a conflict.

Initium-Odp/war/odp/javascript/script.js
Emperorlou: Disabling shortcuts is kind of an old concept for the full page reloads....
I wonder if it would be better to get rid of it altogether...

NikeJoshua:
I didn't really fix anything for the iframe, just cleaned up the code a bit. If it's difficult to sort out, just use whatever he has, and i'll take another look at that part once i update to your latest version after all the merges.

Initium-Odp/war/odp/javascript/script.js
NikeJoshua: Yeah, it's not really used much, except to disable things for the long operations

Also, added a small fix for the use of quietly in refresh.

PR #22, opened 2016-05-13
CommandAttack implementation

NikeJoshua:


Initium-Odp/src/com/universeprojects/miniup/server/commands/CommandAttack.java
```diff
@@ -0,0 +1,200 @@
...
+			canAttackHelper(ds, location, character, monster);
+			return true;
+		} catch (Exception e) {
+			return false;
+		}
+	}
+	
+	private static void canAttackHelper(CachedDatastoreService ds, CachedEntity location, CachedEntity character, CachedEntity monster) throws UserErrorMessage
+	{
+		if (ds==null || location==null || character==null || monster==null)
```
Emperorlou: This is really not very helpful. Better to write out each if statement separately with a proper message so you know what was null. Even an NPE would be better.

Initium-Odp/src/com/universeprojects/miniup/server/commands/CommandAttack.java
```diff
@@ -0,0 +1,200 @@
...
+			throw new UserErrorMessage("You're too busy to attack anything at the moment.");
+		
+		// Additional incapacitated check just in case mode wasn't set properly.
+		if (GameUtils.isPlayerIncapacitated(character))
+			throw new UserErrorMessage("You're incapacitated at the moment.");
+		
+		// Check if monster is alive
+		Double hitpoints = (Double)monster.getProperty("hitpoints");
+		mode = (String)monster.getProperty("mode");
+		if (hitpoints==null || hitpoints<=0 || "DEAD".equals(mode))
```
Emperorlou: Better to do an incapacitated check here as well. isPlayerIncapacitated shouldn't be specific to a player, my bad with the naming.

Initium-Odp/src/com/universeprojects/miniup/server/commands/CommandAttack.java
```diff
@@ -0,0 +1,200 @@
...
+		{
+			// DB sanity check. Mode should be DEAD
+			if ("DEAD".equals(mode)==false)
```
Emperorlou: You know what... maybe instead of having this sort of code all over the place we should have a "boolean GameUtils.normalizeDatabaseState(CachedEntity entity)" call which would do all sorts of different checks to make sure everything makes sense.

Emperorlou: Alternatively it might be even better to make it specific to a kind (to avoid method size bloat). 

So like normalizeDatabaseState_Character(CachedEntity entity)

Initium-Odp/src/com/universeprojects/miniup/server/commands/CommandAttack.java
NikeJoshua: This is only here for the canAttack call really, and there any error is meaningless other than returning false. The 1 specific case where monster could be null from the Attack command is checked before calling the function.

Initium-Odp/src/com/universeprojects/miniup/server/commands/CommandAttack.java
NikeJoshua: The reason for this manual hp check is that incapacitated checks for Zombie status. I assumed that a monster with Zombie set would still make a valid monster, unlike it being a valid player. If this is incorrect and Zombie status makes monsters inelligible to be attacked, then obviously the gameutils call would be better.

Emperorlou: Ahh nuts. Yea you're right and I bet that's why I said isPLAYERIncapacitated() too. Maybe we should have a isMonsterIncapacitated() then.

Initium-Odp/src/com/universeprojects/miniup/server/commands/CommandAttack.java
Emperorlou: Oh I see what you're saying. Okay!

PR #60, opened 2016-08-03
Group changes

Atmostphear:


Atmostphear:
If someone could test these changes out and let me know what I need to fix that would be great.

PR #78, opened 2016-08-08
Group delete command & bug fixes

Atmostphear:
Added a new command to delete a group. Fixed some bugs with other group related stuff.

### DIFF
--- a/Initium-Odp/src/com/universeprojects/miniup/server/commands/CommandGroupDelete.java
+++ b/Initium-Odp/src/com/universeprojects/miniup/server/commands/CommandGroupDelete.java
@@ -58,7 +58,7 @@ public class CommandGroupDelete extends Command
 		if (members.size() > 1)
 		{
 			throw new UserErrorMessage(
-					"The group must be empty before it can be deleted.");
+					"The group must have no members before it can be deleted.");
 		}
 
 		if (((Key) group.getProperty("creatorKey")).getId() != character
--- a/Initium-Odp/src/com/universeprojects/miniup/server/commands/CommandGroupDelete.java
+++ b/Initium-Odp/src/com/universeprojects/miniup/server/commands/CommandGroupDelete.java
@@ -1,0 +1,65 @@
+package com.universeprojects.miniup.server.commands;
+
+import java.util.Map;
+
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+
+import com.google.appengine.api.datastore.Key;
+import com.universeprojects.cacheddatastore.CachedDatastoreService;
+import com.universeprojects.cacheddatastore.CachedEntity;
+import com.universeprojects.miniup.server.ODPDBAccess;
+import com.universeprojects.miniup.server.commands.framework.Command;
+import com.universeprojects.miniup.server.commands.framework.UserErrorMessage;
+
+/**
+ * Delete group command.
+ * 
+ * @author Atmostphear
+ */
+
+public class CommandGroupDelete extends Command
+{
+
+	/**
+	 * Command to delete a group.
+	 * 
+	 * @param request
+	 *            Server request
+	 * @param response
+	 *            Server response
+	 */
+	public CommandGroupDelete(final ODPDBAccess db,
+			final HttpServletRequest request, final HttpServletResponse response)
+	{
+		super(db, request, response);
+	}
+
+	@Override
+	public final void run(final Map<String, String> parameters)
+			throws UserErrorMessage
+	{
+		// Variables
+		ODPDBAccess db = getDB();
+		CachedDatastoreService ds = getDS();
+		CachedEntity character = db.getCurrentCharacter(request);
+		Key groupKey = (Key) character.getProperty("groupKey");
+		CachedEntity group = db.getEntity(groupKey);
+		String groupName = (String) group.getProperty("name");
+
+		// Deleting the group
+		group.setProperty("createdDate", null);
+		group.setProperty("creatorKey", null);
+		group.setProperty("name", null);
+		group.setProperty("applicationMode", null);
+
+		ds.put(group);
+
+		// Removing character from group
+		db.doLeaveGroup(ds, character);
+
+		ds.put(character);
+
+		setPopupMessage(groupName + " has been deleted.");
+	}
+}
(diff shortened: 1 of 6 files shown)


Initium-Odp/src/com/universeprojects/miniup/server/commands/CommandGroupDelete.java
```diff
@@ -0,0 +1,85 @@
...
+		String groupName = (String) group.getProperty("name");
+		List<CachedEntity> members = db.getGroupMembers(null, group);
+
+		if (members.size() > 1)
+		{
+			throw new UserErrorMessage(
+					"The group must have no members before it can be deleted.");
+		}
+
+		if (((Key) group.getProperty("creatorKey")).getId() != character
```
Emperorlou: This should instead use GameUtils.equals(Key,Key)

Initium-Odp/src/com/universeprojects/miniup/server/commands/CommandGroupDelete.java
```diff
@@ -0,0 +1,85 @@
...
+				.getKey().getId())
+		{
+			throw new UserErrorMessage("Only the creator can delete the group.");
+		}
+
+		// Deleting the group
+		group.setProperty("createdDate", null);
```
Emperorlou: All these  (including the ds.put()) is unnecessary. The proper way to delete the group is ds.delete(). There is no need to update anything on the group since you're deleting it anyway.

Initium-Odp/src/com/universeprojects/miniup/server/commands/CommandGroupDelete.java
```diff
@@ -0,0 +1,85 @@
...
+		group.setProperty("creatorKey", null);
+		group.setProperty("name", null);
+		group.setProperty("applicationMode", null);
+
+		ds.put(group);
+
+		// Removing character from group
+		db.doLeaveGroup(ds, character);
```
Emperorlou: I'm not suuure if this will work, it will probably error out because the group is deleted at this point. Better to just nullify the character's groupKey, groupStatus, and groupRank properties in stead.

Initium-Odp/src/com/universeprojects/miniup/server/commands/CommandGroupMemberDemoteFromAdmin.java
```diff
@@ -70,6 +70,12 @@ public final void run(final Map<String, String> parameters)
 					"You are not the creator of your group and cannot perform this action.");
 		}
 
+		if (demoteCharacter.getKey().equals(group.getProperty("creatorKey")))
```
Emperorlou: This should use GameUtils.equals(Key,Key)

Initium-Odp/src/com/universeprojects/miniup/server/commands/CommandGroupMemberKick.java
```diff
@@ -67,7 +67,12 @@ public final void run(final Map<String, String> parameters)
 		if (!"Admin".equals(admin.getProperty("groupStatus")))
```
Emperorlou: As a convention preference, I prefer to use ==false because it can be easy to miss a little !. It also reads better. Readability is highly preferred.

Initium-Odp/src/com/universeprojects/miniup/server/commands/CommandGroupMemberKick.java
```diff
@@ -67,7 +67,12 @@ public final void run(final Map<String, String> parameters)
...
 		{
 			throw new UserErrorMessage(
-					"You are not an admin of your group and cannot perform this action");
+					"You are not an admin of your group and cannot perform this action.");
+		}
+
+		if (kickCharacter.getKey().equals(group.getProperty("creatorKey")))
```
Emperorlou: GameUtils.equals(key,key)

Initium-Odp/war/odp/ajax_group.jsp
```diff
@@ -249,6 +249,10 @@
 					out.println("<a onclick='groupMemberKickCancel("
 							+ character.getKey().getId()
 							+ ")'>Cancel Kick</a>");
+				if (((Key) group.getProperty("creatorKey")).getId() == character
```
Emperorlou: GameUtils.equals(Key,Key)

Initium-Odp/war/odp/javascript/script.js
```diff
@@ -759,6 +759,14 @@ function createNewGroup(eventObject)
 }
 
 
+function deleteGroup(eventObject)
+{
+	confirmPopup("Confirmation", "Are you sure you want to delete your group?\n\nThis cannot be undone", function(){
+		doCommand(eventObject, "GroupDelete")
```
Emperorlou: Maybe it would be good to close the page popup in a callback? If you do, make sure to check for errors first.

IIRC, closePagePopup() or something similar

Initium-Odp/src/com/universeprojects/miniup/server/commands/CommandGroupDelete.java
Atmostphear: Currently the only thing the doLeaveGroup method does is set those properties to null but just in case we want to change doLeaveGroup down the line I'll manually null the properties here. I could also rename the doLeaveGroup to doRemoveGroupProperties since there are multiple commands that set the group properties to null.

Emperorlou: Oh I see what you're saying. Yea if doLeaveGroup doesn't actually use or fetch the group entity, then its certainly fine to use. 

As for the rename, sure, though I think leave group sounds ok to me and should be fine to use elsewhere if all it does is nullify properties.

Initium-Odp/src/com/universeprojects/miniup/server/commands/CommandGroupDelete.java
```diff
@@ -0,0 +1,81 @@
...
+		{
+			throw new UserErrorMessage("Only the creator can delete the group.");
+		}
+
+		// Deleting the group
+		ds.delete(group);
+
+		db.doLeaveGroup(ds, character); // Sets group properties to null
+
+		ds.put(character); // This is a bit redundant since it happens in
```
Emperorlou: Actually saving an entity twice in the same request is usually bad news it should be removed. Thanks for the comments!

PR #96, opened 2016-08-22
Fixed TradeRemoveItem

Davos-Onomous:


Initium-Odp/src/com/universeprojects/miniup/server/commands/CommandTradeRemoveItem.java
```diff
@@ -31,6 +32,7 @@ public void run(Map<String,String> parameters) throws UserErrorMessage {
 		Key otherCharacter = (Key) character.getProperty("combatant");
         CachedEntity item = db.getEntity("Item", itemId);
         TradeObject tradeObject = TradeObject.getTradeObjectFor(ds, character);
+        List<CachedEntity> saleItems = db.getSaleItemsFor(character.getKey());
```
Emperorlou: This COULD be done more efficiently. Since we only care about 1 item, we could do a query to fetch the SaleItem for just that one item. If there are none, just pass an empty list into the HtmlComponents method, if there is 1, then add that 1 to a list and pass that list in.

PR #137, opened 2016-09-22
Multi-item processing fixes

SPFiredrake:
Add checkboxes to trade window. Return itemId when selling, to remove from original list. Add override annotation to getEntityType of ItemsStoreDelete, and add all items to processedItems (so they can be removed from the original list).


Initium-Odp/src/com/universeprojects/miniup/server/HtmlComponents.java
```diff
@@ -241,6 +241,7 @@ public static String generateTradeInvItemHtml(CachedEntity item, List<CachedEnti
 		String result = "";
 		   result+="<div class='invItem' ref="+item.getKey().getId()+">";
 		   result+="<div class='main-item'>";
+		   result+="<input type=checkbox>";
```
Emperorlou: I'm sure it'll be ok but proper html formatting is something more like this: <input type='checkbox'/>

PR #140, opened 2016-09-22
Generate new HTML from Item instead of SaleItem.

SPFiredrake:


Initium-Odp/src/com/universeprojects/miniup/server/commands/CommandItemsStoreDelete.java
```diff
@@ -32,19 +34,22 @@ public CommandItemsStoreDelete(ODPDBAccess db, HttpServletRequest request, HttpS
...
+		
 		Key characterKey = character.getKey();
 		StringBuilder storeString = new StringBuilder();
 		for(CachedEntity storeItem:batchItems)
 		{
 			if (GameUtils.equals(characterKey, storeItem.getProperty("characterKey"))==false)
 				continue;
 			
+			// We need the actual item to generate the HTML to add.
+			CachedEntity item = db.getEntity((Key)storeItem.getProperty("itemKey"));
```
Emperorlou: This would be quite a bit more efficient and faster if you did a batch get of all the items you're going to need before entering this loop.

PR #143, opened 2016-09-23
Mitigate sound toggle on settings page not working

Murnto:
`ajax_settings.jsp` already passes the checkbox state as a parameter to `toggleEnvironmentSoundEffects`, so we'll use that to ensure we're always in sync with the checkbox.

**Edit:** Also save arguments to setAudioDescriptor when sound is disabled. So if the user had disabled sound prior to page load, they may actually enable it without needing to reload.

### DIFF
--- a/Initium-Odp/war/odp/javascript/script.js
+++ b/Initium-Odp/war/odp/javascript/script.js
@@ -2705,8 +2705,12 @@ function toggleEnvironmentSoundEffects(newState)
 	
 	createjs.Sound.muted = enabled;
 	localStorage.setItem("checkboxDisableEnvironmentSoundEffects", enabled+"");
-	
-	
+
+	if (requestedAudioDescriptor !== null)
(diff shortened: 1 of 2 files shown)


Emperorlou:
There are actually 2 places where sound effects get toggled. The little speaker icon in the top right of the page has always worked, but it doesn't pass the check state into the function. Does your solution take this into account?

Emperorlou:
It looks like it was made to work with the speaker icon in the top right and the settings page was forgotten about at some point.

Murnto:
Yes, if the parameter is undefined, it will do the old behavior. Which in awkward javascript replacement tests, worked fine.

https://github.com/Emperorlou/Initium-ODP/pull/143/commits/ff92043241f361e91184658217a9645ddcf4faa3#diff-a3a8a136ef885a2774b74b51a6617361R2702

PR #145, opened 2016-09-23
Migrate buyHouse() to a command

Murnto:


Emperorlou:
Great. You'll need to update the function that triggers the buyHouse in script.js to do a doCommand() call to your command now instead.

PR #146, opened 2016-09-23
Add new Unequip commands. Utilize POST for doCommand.

SPFiredrake:
Fix bug in CommandItemsStoreDelete. Clean up checkbox logic for selection lists (especially after batch commands). Move the delete item X to the right of the item div. Alter styles so full main-item-container div is clickable and checkbox has some more room.


Initium-Odp/war/odp/ajax_managestore.jsp
```diff
@@ -114,7 +114,7 @@
 				</div>
 				<div class='inventory-main-commands'>
 					<div class='command-row'>
-						<label class='command-cell' title='Marks sold storefront items for batch operations.'><input type='checkbox' class='check-group' ref='soldItems'>Select Sold</label>
+						<label class='command-cell' title='Marks sold storefront items for batch operations.'><input type='checkbox' ${if(saleItems.size()==0)"disabled"} class='check-group' ref='soldItems'>Select Sold</label>
```
Emperorlou: Wow. Does this work?? Neat syntax

PR #185, opened 2016-10-14
cancel long op after failed lockcheck

jDyn90:


Initium-Odp/src/com/universeprojects/miniup/server/services/MovementService.java
```diff
@@ -25,8 +25,10 @@ else if (GameUtils.equals(destinationLocationKey, pathToTake.getProperty("locati
 		
 		if (lockCode != null) 
 		{
-			if (checkHasKey(character, (long)lockCode)==false)
+			if (checkHasKey(character, (long)lockCode)==false) {
+				db.getDB().getMC().put("LongOperation-"+character.getKey(), null);
```
Emperorlou: I believe the problem is more a UI state than anything else. I think we simply need to refresh the page.

Emperorlou: So I guess the calling command should do that.

PR #186, opened 2016-10-14
New rediscover houses command; migrate view_profile to ODP

SPFiredrake:


Initium-Odp/src/com/universeprojects/miniup/server/services/PropertiesService.java
```diff
@@ -80,4 +80,18 @@ public void buyHouse(ODPDBAccess db, CachedDatastoreService ds, CachedEntity use
 		}
 		
 	}
+	
+	public void rediscoverHouses(Key userKey)
```
Emperorlou: I just migrated the discoverAllPropertiesFor() method into the ODP. This is what you're looking for.

PR #195, opened 2016-10-22
adding ContainerService, updating transmute

jDyn90:
Transmute container now loses durability on every attempt


Initium-Odp/src/com/universeprojects/miniup/server/commands/CommandTransmuteItems.java
```diff
@@ -54,8 +56,10 @@ public void run(Map<String, String> parameters) throws UserErrorMessage {
 		
 		final List<CachedEntity> materials = db.getFilteredList("Item", "containerKey", FilterOperator.EQUAL, containerKey);
 		
-		if (materials.size() < 2)
-			throw new UserErrorMessage("You must select at least two items to transmute.");
+		if (materials.size() < 2) {
+			cs.doUse(ds, container, 1);
```
Emperorlou: It shouldn't trigger a use here since the transmute didn't actually proceed.

PR #197, opened 2016-10-30
Added item stack handling

thechosenlne:


Initium-Odp/src/com/universeprojects/miniup/server/commands/CommandItemsStackMerge.java
```diff
@@ -0,0 +1,109 @@
...
+	 * @param entity2 - Second entity
+	 * @return true if they can stack", else false
+	 */
+	protected boolean canStack(CachedEntity entity1, CachedEntity entity2){
+		Map<String, Object> entity1Props = entity1.getProperties();
+		Map<String, Object> entity2Props = entity2.getProperties();
+		if (entity1Props.size()!=entity2Props.size()) {return false;} //simplest check
+		// could instead remove quantity and do equal check between the two maps?
+		for (String checking : entity1Props.keySet()) {
+			if (checking != "quantity") {
```
Emperorlou: In java, all string comparisons must be done using .equals(). Otherwise you're just comparing references.

Initium-Odp/src/com/universeprojects/miniup/server/commands/CommandItemsStackMerge.java
```diff
@@ -0,0 +1,109 @@
...
+				if (!entity1Props.get(checking).equals(entity2Props.get(checking))){
```
Emperorlou: I recommend using GameUtils.equals() for checking equality of entity field values. It handles keys in a special way that makes it compatible with remote API (something that I use when testing the full site locally).

PR #207, opened 2016-11-27
Updates the chat help with some missing commands.

tacobowl8:
Commands added: /faq /guide /wiki /group and /groups


Emperorlou:
I assume you reformatted the entire js file. This makes it so I can't see the differences so I'll have to get you to resubmit with just your code changes. I also use whitespace to separate the different sections to make it easier to find stuff when just browsing (though I didn't actually check if you kept those in or not). It's a good thought, but (and you're not the first ^^) hopefully it won't be too difficult to resubmit with just the help changes.

tacobowl8:
Weird... I'll try resubmitting it.

PR #212, opened 2016-12-04
Added updateEquips to CombatService

thechosenlne:
Removes all items on the current character for which the strength requirement is no longer met (i.e. a buff is applied or expires).

### DIFF
--- a/Initium-Odp/src/com/universeprojects/miniup/server/services/CombatService.java
+++ b/Initium-Odp/src/com/universeprojects/miniup/server/services/CombatService.java
@@ -40,11 +40,10 @@ public class CombatService extends Service
 	}
 	
 	/**
-	 * Removes all items on the current character for which the strength requirement
+	 * Removes all items from the specified character for which the strength requirement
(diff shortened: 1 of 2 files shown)


Initium-Odp/src/com/universeprojects/miniup/server/services/CombatService.java
```diff
@@ -37,4 +38,27 @@ public void enterCombat(CachedEntity attacker, CachedEntity defender, boolean au
...
 		db.sendNotification(ds, defender.getKey(), NotificationType.fullpageRefresh);
 	}
-}
+	
+	/**
+	 * Removes all items on the current character for which the strength requirement
+	 * is no longer met (i.e. a buff is applied or expires).
+	 */
+	public void updateEquips(){
+		CachedEntity character = db.getCurrentCharacter();
```
Emperorlou: Actually I just realized in case of PvP, we will need to be able to specify a character (since the defender in a pvp scenario never actually issues his own server requests, he is merely defending automatically)

PR #213, opened 2016-12-04
Added updateEquips to CombatService

thechosenlne:
Removes all items from a specified character for which the strength requirement is no longer met (i.e. a buff is applied or expires).


Initium-Odp/src/com/universeprojects/miniup/server/services/CombatService.java
```diff
@@ -64,4 +65,26 @@ public boolean isInCombat(CachedEntity character)
...
+	/**
+	 * Removes all items from the specified character for which the strength requirement
+	 * is no longer met (i.e. a buff is applied or expires).
+	 */
+	public void updateEquips(CachedEntity character){
+		CachedDatastoreService ds = db.getDB();
+		for(String slot:ODPDBAccess.EQUIPMENT_SLOTS)
+		{
+			Key itemKey = (Key)character.getProperty("equipment"+slot);
+			CachedEntity item = db.getEntity(itemKey);
```
Emperorlou: It's better to do a batch get instead of hitting the server separately for each piece.

Emperorlou:
(Note to myself) Waiting for @thechosenlne to make the code changes after the latest review.

PR #214, opened 2016-12-04
Allows parties to enter group/player housing if all members of that g…

tacobowl8:
…roup have access by default to that housing.

### DIFF
--- a/Initium-Odp/src/com/universeprojects/miniup/server/ODPDBAccess.java
+++ b/Initium-Odp/src/com/universeprojects/miniup/server/ODPDBAccess.java
@@ -4755,6 +4755,8 @@ public class ODPDBAccess
 		if (destination.getProperty("ownerKey")!=null)
 		{
 			if (isInParty) {
+				//We check to see if all members of the party have access by default to enter
+				//the owned housing.  If not, we reject.
 				List<CachedEntity> partyMembers = getParty(db, character);
 				List<CachedEntity> partyGroups = new ArrayList<CachedEntity>();
 				List<CachedEntity> partyUsers = new ArrayList<CachedEntity>();
@@ -4762,7 +4764,7 @@ public class ODPDBAccess
 					//Removes those who have just applied
 					String groupStatus = (String)character.getProperty("groupStatus");
 					if(("Member".equals(groupStatus)==false && "Admin".equals(groupStatus)==false)) {
-						partyGroups.add(new CachedEntity());
+						partyGroups.add(null);
 					} else {
 						partyGroups.add(partyMember.getProperty("groupKey"));
 					}
@@ -4774,13 +4776,12 @@ public class ODPDBAccess
 					//if there are any differences, set the flag to false
 					boolean allInSameGroup = true;
 					for(int i = 0; i < partyGroup.size() - 1; i++) {
-						if(!partyGroups.get(i).equals(partyGroups.get(i+1))){
+						if(!GameUtils.equals(partyGroups.get(i), partyGroups.get(i+1))){
 							allInSameGroup = false;
+							break;
 						}
 					}
-					if(allInSameGroup && ownerKey.equals(partyGroups.get(0))) {
-						return destination;
-					} else {
+					if(!allInSameGroup && !GameUtils.equals(ownerKey, partyGroups.get(0))) {
 						throw new UserErrorMessage("You cannot enter a group owned house in a party unless all members of the party are part of that group.");
 					}
 				} else if(ownerKey.getKind().equals("User")) {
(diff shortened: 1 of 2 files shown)


tacobowl8:
Rather, all members of that party have access by default to that housing.

Initium-Odp/src/com/universeprojects/miniup/server/ODPDBAccess.java
```diff
@@ -4754,8 +4754,55 @@ else if (currentLocationKey.getId()==pathLocation2Key.getId())
...
+					//iterates through the list of groups that the party members belong to.
+					//if there are any differences, set the flag to false
+					boolean allInSameGroup = true;
+					for(int i = 0; i < partyGroup.size() - 1; i++) {
+						if(!partyGroups.get(i).equals(partyGroups.get(i+1))){
+							allInSameGroup = false;
+						}
+					}
+					if(allInSameGroup && ownerKey.equals(partyGroups.get(0))) {
+						return destination;
```
Emperorlou: You're returning here, bypassing the rest of the method... is that intentional?

Initium-Odp/src/com/universeprojects/miniup/server/ODPDBAccess.java
```diff
@@ -4754,8 +4754,55 @@ else if (currentLocationKey.getId()==pathLocation2Key.getId())
...
+					} else {
+						throw new UserErrorMessage("You cannot enter a group owned house in a party unless all members of the party are part of that group.");
+					}
+				} else if(ownerKey.getKind().equals("User")) {
+					//iterates through the list of accounts that the party members belong to.
+					//if there are any differences, set the flag to false
+					boolean allOfSameUser = true;
+					for(int i = 0; i < partyUsers.size() - 1; i++) {
+						if(!partyUsers.get(i).equals(partyUsers.get(i+1))){
```
Emperorlou: When comparing entities or keys, use GameUtils.equals(). There is sometimes an issue when comparing keys while I'm debugging locally using Remote API and this equals check fixes this weird issue.

Also, it doesn't mind if either side is null, so that's a bonus.

Initium-Odp/src/com/universeprojects/miniup/server/ODPDBAccess.java
```diff
@@ -4754,8 +4754,55 @@ else if (currentLocationKey.getId()==pathLocation2Key.getId())
...
+							allOfSameUser = false;
+						}
+					}
+					if(allOfSameUser && ownerKey.equals(partyUsers.get(0))) {
```
Emperorlou: GameUtils.equals() again here

Emperorlou: Also... you're comparing CachedEntity objects.. might want to look in the .equals() to make sure it actually just compares keys. If not, you'll want to compare keys yourself instead of the entities.

Initium-Odp/src/com/universeprojects/miniup/server/ODPDBAccess.java
```diff
@@ -4754,8 +4754,55 @@ else if (currentLocationKey.getId()==pathLocation2Key.getId())
...
+					partyUsers.add(partyMember.getProperty("userKey"));
+				}
+				Key ownerKey = (Key)destination.getProperty("ownerKey");
+				if(ownerKey.getKind().equals("Group")){
+					//iterates through the list of groups that the party members belong to.
+					//if there are any differences, set the flag to false
+					boolean allInSameGroup = true;
+					for(int i = 0; i < partyGroup.size() - 1; i++) {
+						if(!partyGroups.get(i).equals(partyGroups.get(i+1))){
+							allInSameGroup = false;
```
Emperorlou: Could break here to stop checking since theres no reason to keep checking once you've found they're not in the same group.

Initium-Odp/src/com/universeprojects/miniup/server/ODPDBAccess.java
```diff
@@ -4754,8 +4754,55 @@ else if (currentLocationKey.getId()==pathLocation2Key.getId())
...
-			if (isInParty)
+			if (isInParty) {
+				List<CachedEntity> partyMembers = getParty(db, character);
+				List<CachedEntity> partyGroups = new ArrayList<CachedEntity>();
+				List<CachedEntity> partyUsers = new ArrayList<CachedEntity>();
+				for(CachedEntity partyMember: partyMembers) {
+					//Removes those who have just applied
+					String groupStatus = (String)character.getProperty("groupStatus");
+					if(("Member".equals(groupStatus)==false && "Admin".equals(groupStatus)==false)) {
+						partyGroups.add(new CachedEntity());
```
Emperorlou: I guess you're creating the entity to avoid a null? If you do GameUtils.equals() calls to check for comparison, you could actually pass it a null and avoid creating new entities that way.

Initium-Odp/src/com/universeprojects/miniup/server/ODPDBAccess.java
tacobowl8: Yes, this is intentional.  We already know the destination is someplace that has an owner, and we have done all of the checks that they should make to be able to enter, so returning here should be fine.

Initium-Odp/src/com/universeprojects/miniup/server/ODPDBAccess.java
tacobowl8: I was considering that, but with the max party size being 4, I got a little lazy.  I can definitely look into doing that.

Initium-Odp/src/com/universeprojects/miniup/server/ODPDBAccess.java
Emperorlou: I had a closer look and I think key checks are done aftewards. We should definitely run through that first (and any other checks we may add in the future).

PR #219, opened 2016-12-10
Changed datatypes from double to long

zbutts1992:


Initium-Odp/src/com/universeprojects/miniup/server/GameUtils.java
```diff
@@ -884,8 +884,8 @@ public static String renderItem(ODPDBAccess db, HttpServletRequest request, Cach
 			if (strengthRequirement!=null && characterStrength<strengthRequirement)
 				hasRequiredStrength = false;
 			
-			double maxDura = (Double)item.getProperty("maxDurability");
-			double currentDura = (Double)item.getProperty("durability");
+			long maxDura = (Long)item.getProperty("maxDurability");
```
Emperorlou: It's possible for an item to have null dura, so long wont work. Needs to be Long.

PR #225, opened 2016-12-12
A fix and a test

Malzawar:
Fixed an always thrown UserErrorMessage.

Testing how to properly find duplicate items.


Emperorlou:
Looks great and this will work. One thing though, it's better if you use a new feature I implemented recently to automatically take all .put() calls and instead enter them in bulk at the end of the command.  Do that with the following:

ds.beginBulkWriteMode()

// Put all your .put() calls in between

ds.commitBulkWrite();

You can begin the bulk write mode at the very start of the command and then commit at the end. This is an optimization that will speed up your command execution considerably by making 1 round trip instead of nearly 20.

PR #241, opened 2016-12-14
CommandItemsStorageContainerSwap

RevMuun:
Ready for code review.

### DIFF
--- a/Initium-Odp/src/com/universeprojects/miniup/server/commands/CommandItemsSwapStorageContainers.java
+++ b/Initium-Odp/src/com/universeprojects/miniup/server/commands/CommandItemsSwapStorageContainers.java
@@ -42,37 +42,45 @@ public class CommandItemsSwapStorageContainers extends CommandItemsBase {
 					throw new UserErrorMessage("Character does not have access to container.");
 			}
 			
+			CachedEntity containerOne = batchItems.get(0);
+			CachedEntity containerTwo = batchItems.get(1);
+			List<CachedEntity> itemOneContents = db.getItemContentsFor(containerOne.getKey());
+			List<CachedEntity> itemTwoContents = db.getItemContentsFor(containerTwo.getKey());
+			
 			//check that at least one container is empty.
-			if(db.getItemContentsFor(batchItems.get(0).getKey()).size() != 0 && 
-					db.getItemContentsFor(batchItems.get(1).getKey()).size() != 0){
+			if(itemOneContents.size() != 0 && 
+					itemTwoContents.size() != 0){
 				throw new UserErrorMessage("At least one storage item must be empty in order to swap contents.");
 			}				
 			
+					
 			//determine which container is empty. Since we checked that at least one container is empty,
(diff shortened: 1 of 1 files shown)


Initium-Odp/src/com/universeprojects/miniup/server/commands/CommandItemsSwapStorageContainers.java
```diff
@@ -0,0 +1,83 @@
...
+			if(db.getItemContentsFor(batchItems.get(0).getKey()).size() != 0 && 
+					db.getItemContentsFor(batchItems.get(1).getKey()).size() != 0){
+				throw new UserErrorMessage("At least one storage item must be empty in order to swap contents.");
+			}				
+			
+			//determine which container is empty. Since we checked that at least one container is empty,
+			//we can assume that if index 1 has equipment, then index 0 is empty.
+			int emptyContainerIndex = 0;
+			int fullContainerIndex = 1;
+			if(db.getItemContentsFor(batchItems.get(1).getKey()).size() == 0){
```
Emperorlou: Each of these calls hit the database separately. Better cache the results since you use it more than once.

Initium-Odp/src/com/universeprojects/miniup/server/commands/CommandItemsSwapStorageContainers.java
```diff
@@ -0,0 +1,83 @@
...
+			Long spaceReq = db.getItemCarryingSpace(batchItems.get(fullContainerIndex));
+			Long weightReq = db.getItemCarryingWeight(batchItems.get(fullContainerIndex));
+			Long spaceAvail = (Long) batchItems.get(emptyContainerIndex).getProperty("maxSpace");
+			Long weightAvail = (Long) batchItems.get(emptyContainerIndex).getProperty("maxWeight");
+			
+			//check that we have room in the new container to fit everything.
+			if(spaceAvail < spaceReq)
+				throw new UserErrorMessage("There is not enough space available in the new container.");
+			if(weightAvail < weightReq)
+				throw new UserErrorMessage("There is not enough weight available in the new container.");
```
Emperorlou: This error message is player facing so should be a little less robotic. Something like "the container you're swapping in to cannot carry so much weight. Try a different container perhaps."

Initium-Odp/src/com/universeprojects/miniup/server/commands/CommandItemsSwapStorageContainers.java
```diff
@@ -0,0 +1,83 @@
...
+			
+			//actually move the content now and save to database.
+			ds.beginBulkWriteMode();
+			for(CachedEntity item : db.getItemContentsFor(batchItems.get(fullContainerIndex).getKey())){
```
Emperorlou: For code readability, rather than calling get using the index, it might be better to actually cache the entity instead. Minor of course.

RevMuun:
@Emperorlou I applied your advice and have pushed a new commit.

PR #251, opened 2016-12-14
Removed debug messages from container swap.

RevMuun:


RevMuun:
Pushed a commit with a more explicit message.

RevMuun:
@Emperorlou done. Added helper function to ContainerService and updated confirmation message.

PR #254, opened 2016-12-14
Alliance war commands

BrawlingKhan:
Let me know of any issues


BrawlingKhan:
I gotcha, my jsp was also messed up.

PR #256, opened 2016-12-15
Command Migration] Make the "drink beer" into a proper command

Fayyne:
Let me know if anything seems off, I dug around other commands, and look around in what I was calling to make sure messages would still prop. as needed, and I think I got it, but still learning the process flow of the app.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,8 @@
 /Initium-Odp/war/WEB-INF/lib/appengine-endpoints.jar
 /Initium-Odp/war/WEB-INF/lib/appengine-endpoints-deps.jar
 /Initium-Odp/war/WEB-INF/lib/appengine-jsr107cache-1.9.36.jar
+*.iml
+*.eml
+.idea/
+.classpath
+/Initium-Odp/.classpath
--- a/.gitignore
+++ b/.gitignore
@@ -9,5 +9,3 @@
 *.iml
 *.eml
 .idea/
-.classpath
-/Initium-Odp/.classpath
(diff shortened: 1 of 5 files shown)


.gitignore
```diff
@@ -6,3 +6,8 @@
 /Initium-Odp/war/WEB-INF/lib/appengine-endpoints.jar
 /Initium-Odp/war/WEB-INF/lib/appengine-endpoints-deps.jar
 /Initium-Odp/war/WEB-INF/lib/appengine-jsr107cache-1.9.36.jar
+*.iml
+*.eml
+.idea/
+.classpath
```
Emperorlou: I don't want to ignore the class path so new devs have something to work with, even if it's wrong.

Initium-Odp/war/odp/javascript/script.js
```diff
@@ -1547,11 +1547,12 @@ function deleteAndRecreateCharacter(currentCharName)
 	});
 }
 
-function doDrinkBeer()
+function doDrinkBeer(eventObject)
 {
 	confirmPopup("Drink Beer", "Are you sure you want to drink? It might affect your combat abilities...", function(){
```
Emperorlou: Maybe we shouldn't warm them... For fun ^^

.gitignore
Fayyne: FUDGE didn't mean to put that one in there, habit >.>

Initium-Odp/war/odp/javascript/script.js
```diff
@@ -1547,11 +1547,12 @@ function deleteAndRecreateCharacter(currentCharName)
...
-		window.location.href = "ServletCharacterControl?type=drinkBeer"+"&v="+window.verifyCode;
-		
+        closeAllPopups();
```
Emperorlou: Let's not close pop-ups or tooltips anymore too

PR #258, opened 2016-12-15
Changed to one command for declare and end

BrawlingKhan:


Initium-Odp/war/WEB-INF/odppages/ajax_group.jsp
```diff
@@ -6,6 +6,16 @@
 		</h1>
 		<p>
 			<c:out value="${groupDescription}" />
+			<br />
+			<a onclick='declareWar(event)'>Declare War</a>
+			<br />
+			<c:if test="${not empty warDecGroupNames}">
+			<c:forEach var="declaredGroups" items="${warDecGroupNames}">
+				<p>${declaredGroups}</p>
+				<a onclick='declareWar(event, "<c:out value="${declaredGroups}"/>")'>End War</a>
```
Emperorlou: We will only want to show this option if the user is an admin

PR #262, opened 2016-12-15
Small fix for an NPE

Malzawar:


Initium-Odp/src/com/universeprojects/miniup/server/commands/CommandCharacterEquipSet.java
```diff
@@ -98,7 +98,7 @@ public int compare(CachedEntity e1, CachedEntity e2) {
 
 		for (CachedEntity equipment : setEquip) {
 
-			if (slotList.size() == 0) {
+			if (slotList.size() == 0 || slotList==null) {
```
Emperorlou: Null check should be first or this will throw an npe

Malzawar: Actually, it never is going to be null at that point, since I initialized it before. That is a leftover of a thinking process, sorry. Will remove.

PR #278, opened 2016-12-18
Adds a forget all combat sites link if there are atleast two forgettable combat sites on the page.

tacobowl8:


Initium-Odp/src/com/universeprojects/miniup/server/commands/CommandForgetAllCombatSites.java
```diff
@@ -0,0 +1,72 @@
...
+		
+		//The location the command is being called from
+		CachedEntity location = db.getLocationById(tryParseId(parameters, "locationId"));
+		
+		if(GameUtils.equals(location.getKey(), characterLocationKey))
+			throw new RuntimeException("Player is not at the location they are forgetting combat sites from.");
+		
+		//We first get all the discoveries for the character and the character's current location.
+		List<CachedEntity> discoveries = db.getDiscoveriesForCharacterAndLocation(character.getKey(), characterLocationKey);
+		ds.beginBulkWriteMode();
```
Emperorlou: Might as well put the write mode right at the start, in case some of the get calls actually write to the DB.

tacobowl8: Makes sense!  I'll do that.

PR #286, opened 2016-12-19
Added containsEquippable(container) method to ContainerService

Malzawar:
, that returns true if the container has an equipable item. And false
otherwise.


Initium-Odp/src/com/universeprojects/miniup/server/services/ContainerService.java
```diff
@@ -107,4 +107,18 @@ public boolean containsAll(CachedEntity container, List<CachedEntity> items) {
 		return content.containsAll(items);
 	}
 	
+	public boolean containsEquippable(CachedEntity container){
```
Emperorlou: Definitely give me the option to pass in the container's items in a list if I already have them. If it's null, you can query for them then.

PR #290, opened 2016-12-19
Adds chat help in an html file so that we can bring it up differently.

tacobowl8:


Initium-Odp/war/odp/chatHelp.html
```diff
@@ -0,0 +1,71 @@
+<div id="popup_1" class="popup" >
```
Emperorlou: Just remove this top div and you're good

PR #294, opened 2016-12-21
Adds a duration check so that it will merely end early instead of get…

tacobowl8:
…ting a timeout.


Initium-Odp/src/com/universeprojects/miniup/server/commands/CommandForgetAllCombatSites.java
```diff
@@ -43,7 +46,10 @@ public void run(Map<String, String> parameters) throws UserErrorMessage {
 
 		
 		for(Long forgettableCombatSite : forgettableCombatSiteList) {
+			if(System.currentTimeMillis() - startTime >= MAX_MILLISECONDS_TO_SPEND)
+				break;
```
Emperorlou: Should add a message to the user stating how many sites were removed and how many were still left.

Actually maybe we should stop it at 10 seconds now that I think of it. People probably wouldn't wait for longer than that anyway.

PR #295, opened 2016-12-21
Commandizes the show hidden sites link.  Holding off on replacing tha…

tacobowl8:
…t so I can test it via html modification.

### DIFF
--- a/Initium-Odp/src/com/universeprojects/miniup/server/services/MainPageUpdateService.java
+++ b/Initium-Odp/src/com/universeprojects/miniup/server/services/MainPageUpdateService.java
@@ -81,12 +81,24 @@ public class MainPageUpdateService extends Service
 		this.pathEnds = pathEnds;
 		this.hasHiddenPaths = hasHiddenPaths;
 	}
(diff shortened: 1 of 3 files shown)


Initium-Odp/src/com/universeprojects/miniup/server/services/MainPageUpdateService.java
```diff
@@ -197,7 +209,7 @@ else if (currentLocationKey.getId()==pathLocation2Key.getId())
 						if ("FromLocation2Only".equals(forceOneWay) && currentLocationKey.getId() == pathLocation1Key.getId())
 							continue;
 				
-						if ("TRUE".equals(discovery.getProperty("hidden")) && db.getRequest().getParameter("showHiddenPaths")==null)
+						if ("TRUE".equals(discovery.getProperty("hidden")) && (db.getRequest().getParameter("showHiddenPaths")==null || !showHidden))
```
Emperorlou: You can get rid of the request parameter check that does the same thing

PR #303, opened 2016-12-22
Moved tryParseStringToList method from CommandForgetAllCombatSites to…

tacobowl8:
… Command.

### DIFF
--- a/Initium-Odp/src/com/universeprojects/miniup/server/commands/framework/Command.java
+++ b/Initium-Odp/src/com/universeprojects/miniup/server/commands/framework/Command.java
@@ -73,8 +73,8 @@ public abstract class Command extends OperationBase
 	{
 		try {
 			return Long.parseLong(params.get(fieldName));
-		} catch (Exception _) {
-			throw new RuntimeException(this.getClass().getSimpleName()+" invalid call format, '"+fieldName+"' is not a valid id.");
+		} catch (Exception e) {
+			throw new RuntimeException(this.getClass().getSimpleName()+" invalid call format, '"+fieldName+"' is not a valid id.", e);
 		}
 	}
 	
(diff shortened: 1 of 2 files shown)


Initium-Odp/src/com/universeprojects/miniup/server/commands/framework/Command.java
```diff
@@ -75,6 +77,29 @@ protected Long tryParseId(Map<String,String> params, String fieldName)
...
+		try {
+			String unparsedString = params.get(fieldName);
+			String[] parsedArray = unparsedString.split(delimitingCharacter);
+			ArrayList<Long> parsedList = new ArrayList<Long>();
+			for(int i = 0; i < parsedArray.length; i++) {
+				parsedList.add(Long.parseLong(parsedArray[i]));
+			}
+			return parsedList;
+		} catch (Exception _) {
+			throw new RuntimeException(this.getClass().getSimpleName()+" invalid call format, '"+fieldName+"' is not a valid delimited list.");
```
Emperorlou: Should pass the exception in the re throw so we get the full stack trace

Initium-Odp/src/com/universeprojects/miniup/server/commands/framework/Command.java
```diff
@@ -75,6 +77,29 @@ protected Long tryParseId(Map<String,String> params, String fieldName)
...
+	
+	/** 
+	 * Tries to fetch fieldName from param list and throws if it cannot.
+	 * 
+	 * @param params
+	 * @param fieldName
+	 * @param delimitingCharacter
+	 * @return List of longs, parsed from param string
+	 */
+	protected List<Long> tryParseStringToList(Map<String,String> params, String fieldName, String delimitingCharacter)
```
Emperorlou: It's a bit odd to do this only for a list of Long's without indicating it is for longs in the method name. Overloads will be needed for other types and string is the most obvious candidate for the non specific method name like this. Maybe just add the word Long before List?

Emperorlou: tryParseStringToLongList

PR #304, opened 2016-12-22
Debugging jsp output

PoolRequest:


Initium-Odp/war/WEB-INF/odppages/ajax_group.jsp
```diff
@@ -11,7 +11,7 @@
 			<c:forEach var="groups" items="${warDecGroupNames}">
 				<p>${$groups}</p>
 				<br/>
-				<a onclick='endWar(event, "<c:out value="${groups}"/>")'>End War</a>
+				<a onclick='endWar(event, ${groups})'>End War</a>
```
Emperorlou: This change seems wrong. No quotes around the string argument mostly. Also using c out will escape characters for you.

PoolRequest:
Yea my bad, switched it back.  Command works fine, having trouble getting the output correct though.

PR #310, opened 2016-12-29
Commandize Join/LeaveParty + Some minor neatness

DylanHolmesDH:
I commandized the Join/LeaveParty functions.

The only thing missing is the JavaScript link for joining a party.

### DIFF
--- a/Initium-Odp/src/com/universeprojects/miniup/server/ODPDBAccess.java
+++ b/Initium-Odp/src/com/universeprojects/miniup/server/ODPDBAccess.java
@@ -2936,6 +2936,35 @@ public class ODPDBAccess
 	////////// END GROUP METHODS //////////
 	///////////////////////////////////////
 	
+	/**
+	 * 
+	 * @param db
+	 * @param character, the entity to leave the party
+	 * Takes in an entity and sets their partyCode and partyLeader properties to null (safety checks are done in CommandLeaveParty.java)
+	 */
+	public void doLeaveParty(CachedDatastoreService db, CachedEntity character) {
+		if (ds == null)
+		{
+			ds = getDB();
+		}
+
+		character.setProperty("partyCode", null);
+		character.setProperty("partyLeader", null);
+
+		ds.put(character);
+	}
+	
+	public void doJoinParty(CachedDatastoreService ds2, CachedEntity character, String toJoin) {
+		if (ds == null)
+		{
+			ds = getDB();
+		}
+		
+		character.setProperty("partyCode", toJoin);
+		character.setProperty("partyLeader", "FALSE");
+		
+		ds.put(character);
+	}
 	
 	public void doCharacterDiscoverEntity(CachedDatastoreService db, CachedEntity character, CachedEntity entityToDiscover)
 	{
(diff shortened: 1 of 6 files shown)


Initium-Odp/src/com/universeprojects/miniup/server/commands/CommandPartyJoin.java
```diff
@@ -0,0 +1,56 @@
...
+		if (currentParty != null && !currentParty.trim().equals("")) {
+			throw new UserErrorMessage("You are already in a party! You must leave this one first!");
+		}
+		
+		CachedEntity leader = db.getPartyLeader(ds, toJoin);
+		if (leader.getProperty("partyJoinsAllowed").equals("FALSE")) {
+			throw new UserErrorMessage("This party is not accepting members currently!");
+		}
+		
+		if (character.getProperty("mode").equals("COMBAT")) {
```
Emperorlou: This can cause an npe since the mode can legitimately be null which means normal.

Initium-Odp/src/com/universeprojects/miniup/server/commands/CommandPartyJoin.java
```diff
@@ -0,0 +1,56 @@
...
+			HttpServletResponse response) {
+		super(db, request, response);
+	}
+
+	@Override
+	public void run(Map<String, String> parameters) throws UserErrorMessage {
+		ODPDBAccess db = getDB();
+		CachedDatastoreService ds = getDS();
+
+		CachedEntity character = db.getCurrentCharacter();
```
Emperorlou: Party code will be null when not in a party. We should throw a meaningful error message.

Emperorlou: It is not uncommon to have a stale client screen see we should handle the null situations gracefully.

Emperorlou: Uhhhhhhhhh oops. Ignore everything I said here. I misread the code. ^^

Initium-Odp/src/com/universeprojects/miniup/server/ODPDBAccess.java
```diff
@@ -2949,7 +2978,30 @@ public String cleanCharacterName(String name)
...
+	
+	/**
+	 * Returns the CachedEntity object representing the party leader of @partyCode
+	 * Will return null if the partyCode is empty or null (or if for some reason no one in the party is a leader).
+	 * 
+	 * @param ds
+	 * @param partyCode code of the party to grab the leader from
+	 * @return the party leader of the party.
+	 */
+	public CachedEntity getPartyLeader(CachedDatastoreService ds, String partyCode) {
```
Emperorlou: We should pass in the members of they party as an option, only fetching if the list given is null. Database optimization.

Initium-Odp/src/com/universeprojects/miniup/server/ODPDBAccess.java
```diff
@@ -2949,7 +2978,30 @@ public String cleanCharacterName(String name)
...
+		if (partyCode == null || partyCode.trim().equals("")) {
+			return null;
+		}
+		
+		List<CachedEntity> members = getParty(ds, partyCode);
+		
+		for (CachedEntity member : members) {
+			if (member.getProperty("partyLeader").equals("TRUE")) {
```
Emperorlou: This is a npe

Initium-Odp/src/com/universeprojects/miniup/server/commands/CommandPartyJoin.java
```diff
@@ -0,0 +1,56 @@
...
+		CachedEntity leader = db.getPartyLeader(ds, toJoin);
+		if (leader.getProperty("partyJoinsAllowed").equals("FALSE")) {
+			throw new UserErrorMessage("This party is not accepting members currently!");
+		}
+		
+		if (character.getProperty("mode").equals("COMBAT")) {
+			throw new UserErrorMessage("You cannot join a party while in combat!");
+		}
+		
+		db.doJoinParty(ds, character, toJoin);
```
Emperorlou: I think it would be better to just collapse this call into the command. I can't think of any reason why we would want to call doJoinparty() from anywhere else.

DylanHolmesDH:
@Emperorlou Updated the PR.

PR #318, opened 2016-12-31
First 2D attempt

Dias-Flac:
This is a first attempt at generating grass using simple seeded random numbers and varying height and width. Currently hijacking merchantList jsp to quickly integrate into the env.

### DIFF
--- a/Initium-Odp/war/odp/twoDimSandbox.jsp
+++ b/Initium-Odp/war/odp/twoDimSandbox.jsp
@@ -15,7 +15,7 @@ Seed: <input type='text' id='seed' value=123456 />
 <script>
     $(document).on("click", "#somebutton", function() {
         $.ajax({
-            url: "MyServlet",
+            url: "SandboxServlet",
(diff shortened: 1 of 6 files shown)


Initium-Odp/war/odp/twoDimSandbox.jsp
```diff
@@ -0,0 +1,50 @@
...
+<script>
+    $(document).on("click", "#somebutton", function() {
+        $.ajax({
+            url: "SandboxServlet",
+            data:{width:$("#width").val(), height:$("#height").val(), seed:$("#seed").val()},
+            type: 'POST',
+            success: function(responseJson) {
+
+                // Remove all current images
+                var images = document.getElementsByTagName('img');
```
Emperorlou: This map will be displayed a
Inside the current game page in place of the banner. This JavaScript would make all images disappear, not just the ones in the map I think. You might also consider doing this the jQuery way since it's super nice for this kind of stuff.

PR #320, opened 2016-12-31
Alliance/War work

PoolRequest:
Still gotta put in some checks and make sure things are removed when the group is deleted, mainly testing this before moving forward.


Initium-Odp/src/com/universeprojects/miniup/server/commands/CommandGroupProcessAllianceReq.java
```diff
@@ -0,0 +1,77 @@
...
+		CachedEntity character = db.getCurrentCharacter();
+		String decision = parameters.get("decision");
+
+		Long groupID = parameters.containsKey("groupId") ? tryParseId(parameters, "groupId") : null;
+		if(groupID == null) throw new RuntimeException("Command missing parameter groupId");
+		
+		CachedEntity group = db.getEntity("Group", groupID);
+		GroupService service = new GroupService(db, character);
+		
+		if (decision == "accept")
```
Emperorlou: This won't work in Java. Needs to be .equals

Initium-Odp/src/com/universeprojects/miniup/server/commands/CommandGroupProcessAllianceReq.java
```diff
@@ -0,0 +1,77 @@
...
+			catch(Exception ex)
+			{
+				if(service.characterHasGroup() == false)
+					throw new UserErrorMessage("Character does not belong to a group!");
+				if(service.isCharacterGroupAdmin() == false)
+					throw new UserErrorMessage("Character is not an admin of the group!");
+				throw ex;
+			}		
+		}
+		else if (decision == "decline")
```
Emperorlou: Same here.

PR #353, opened 2017-01-03
87 - Adding new domain/dao classes

kyle-miller:
This is a first attempt at giving developers a set of Java classes for the Initium-ODP domain. It also provides DAO classes for simple querying/saving of said domain objects.


Initium-Odp/src/com/universeprojects/miniup/server/domain/Collectable.java
```diff
@@ -0,0 +1,135 @@
+package com.universeprojects.miniup.server.domain;
+
+import java.util.List;
+
+import com.google.appengine.api.datastore.Key;
+import com.universeprojects.cacheddatastore.CachedEntity;
+
+//This is a "node" that is used to extract items from the game world. Use it for gathering wood, food (like berries/apples/wheat), stone, iron ore...etc.
```
Emperorlou: All comments should definitely be in proper javadoc format!

Initium-Odp/src/com/universeprojects/miniup/server/dao/OdpDao.java
```diff
@@ -0,0 +1,80 @@
...
+			// Each DAO should have tests ensuring this doesn't happen
+			throw new DaoException(String.format("%s is set up incorrectly", getClass().getName()), e);
+		}
+	}
+
+	protected abstract Logger getLogger();
+
+	public abstract T get(Key key);
+
+	public abstract List<T> get(List<Key> keyList) throws DaoException;
```
Emperorlou: This could just be implemented in the base class I think? Keys include the kind as a string Key.getKind().

Initium-Odp/src/com/universeprojects/miniup/server/dao/OdpDao.java
```diff
@@ -0,0 +1,80 @@
...
+			return all;
+		} catch (ReflectiveOperationException e) {
+			// Each DAO should have tests ensuring this doesn't happen
+			throw new DaoException(String.format("%s is set up incorrectly", getClass().getName()), e);
+		}
+	}
+
+	protected abstract Logger getLogger();
+
+	public abstract T get(Key key);
```
Emperorlou: THis also looks like it could be implemented here. Keys include the kind.

Initium-Odp/src/com/universeprojects/miniup/server/dao/OdpDao.java
```diff
@@ -0,0 +1,80 @@
...
+				cachedEntity = null;
+				getLogger().warning("Retrieved an Entity key that has a null value");
+			}
+			return cachedEntity;
+		} catch (EntityNotFoundException e) {
+			return null;
+		}
+	}
+
+	protected List<CachedEntity> findAllCachedEntities(String kind) {
```
Emperorlou: Definitely can't do this at all. Max fetch should be 1k in most cases.

Initium-Odp/src/com/universeprojects/miniup/server/dao/OdpDao.java
```diff
@@ -0,0 +1,80 @@
...
+import java.util.logging.Logger;
+
+import com.google.appengine.api.datastore.EntityNotFoundException;
+import com.google.appengine.api.datastore.Key;
+import com.universeprojects.cacheddatastore.CachedDatastoreService;
+import com.universeprojects.cacheddatastore.CachedEntity;
+import com.universeprojects.miniup.server.domain.OdpDomain;
+import com.universeprojects.miniup.server.exceptions.DaoException;
+
+public abstract class OdpDao<T extends OdpDomain> {
```
Emperorlou: I think this class could benefit by having many of the QueryHelper methods wrapped in it, returning proper wrapped versions of the result sets.

Initium-Odp/src/com/universeprojects/miniup/server/domain/NPCDef.java
```diff
@@ -0,0 +1,225 @@
...
+
+//Defines how to generate a new NPC for the character entity type.
+public class NPCDef extends OdpDomain {
+	public static final String KIND = "NPCDef";
+
+	public NPCDef() {
+		super(new CachedEntity(KIND));
+	}
+
+	public NPCDef(CachedEntity cachedEntity) {
```
Emperorlou: Imo, this would work better as a static method like NPCDef.wrap(cachedEntity). It makes it clear that a new entity is not being created and it's just wrapping.

Initium-Odp/test/com/universeprojects/miniup/server/dao/CustomOrderDaoTest.java
```diff
@@ -0,0 +1,62 @@
+package com.universeprojects.miniup.server.dao;
```
Emperorlou: I don't think we should have a unit test for each kind. They should all be exactly the same and testing the same underlying stuff...

Initium-Odp/src/com/universeprojects/miniup/server/dao/OdpDao.java
kyle-miller-centurylink: Knowing that, yes. I'd imagine that may work.

PR #354, opened 2017-01-03
Creating method to return HTML when in test environment

Dias-Flac:
Creating method to return HTML when in test environment

### DIFF
--- a/Initium-Odp/src/com/universeprojects/miniup/server/services/MainPageUpdateService.java
+++ b/Initium-Odp/src/com/universeprojects/miniup/server/services/MainPageUpdateService.java
@@ -372,20 +372,22 @@ public class MainPageUpdateService extends Service
 	public String updateTestPanel()
 	{
 		if (db.getRequest().getRequestURL().toString().contains("test")) {
-			return "Width: <input type='text' id='width' value=20 />\n"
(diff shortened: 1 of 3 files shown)


kyle-miller:
You may want to make the width/height/offset/seed ids a little more specific. Just in case this becomes part of a larger page. I don't think this holds up this pull request, though.

Initium-Odp/src/com/universeprojects/miniup/server/services/MainPageUpdateService.java
```diff
@@ -364,6 +364,31 @@ public String updateInBannerCharacterWidget()
...
+					+ "<div id=\"somediv\"></div>\n"
+					+ "\n"
+					+ "<script type=\"text/javascript\" src=\"/odp/javascript/Sandbox.js\"></script>\n"
+					+ "<script>\n"
+					+ "    $(document).on(\"click\", \"#somebutton\", function() {\n"
+					+ "        getTilePos();\n"
+					+ "    });\n"
+					+ "</script>\n";
+		}
+		return "";
```
Emperorlou: Look at how the return is called in other methods and try to mimick it.

Emperorlou:
Should also restore the store button with this pr

PR #442, opened 2017-01-24
Updating server side

Dias-Flac:


Initium-Odp/src/com/universeprojects/miniup/server/model/GridCell.java
```diff
@@ -1,9 +1,18 @@
 package com.universeprojects.miniup.server.model;
 
-public class GridBackground {
+public class GridCell {
 	private String backgroundFile;
+	private int xGridCoord;
```
Emperorlou: It would be less work for updates if you just stored everything in a map to begin with. You could even store it in a JSONObject so that it's ready to go. Especially since you're manually writing a json string in the toString(). The only downside is having to cast in your getters.

Initium-Odp/src/com/universeprojects/miniup/server/model/GridObject.java
```diff
@@ -29,6 +29,94 @@ public GridObject(String key, String fileName, String name, int xGridCoord, int
 		this.height = height;
 	}
 
+	public String getKey() {
```
Emperorlou: Same comment as the GridCell, storing everything in a map or JSONObject I think would be easier long-term.

Initium-Odp/src/com/universeprojects/miniup/server/model/UpdateGridCell.java
```diff
@@ -0,0 +1,44 @@
+package com.universeprojects.miniup.server.model;
+
+public class UpdateGridCell {
```
Emperorlou: If you make the change I suggested for the GridCell class, you could just use that for updates.

Initium-Odp/src/com/universeprojects/miniup/server/model/UpdateGridObject.java
```diff
@@ -0,0 +1,34 @@
+package com.universeprojects.miniup.server.model;
```
Emperorlou: Same here.

PR #451, opened 2017-01-25
Adds /help as a chat command.

tacobowl8:


Initium-Odp/war/odp/javascript/messager-impl.js
```diff
@@ -112,6 +112,9 @@ messager.onChatMessage = function(chatMessage)
 			chatMessage.message = "/me is a spammer. Kill him with fire!";
 		var meMode = chatMessage.message.toLowerCase().startsWith("/me ");
 		html+="<span class='chatMessage-text'>";
+		if(chatMessage.message.toLowerCase() == "/help") {
+			chatMessage.message = "<a href=\"#\" onclick=\"pagePopup('/odp/chatHelp.html')\">[Click here to view chat help]</a>";
```
Emperorlou: No I think we should just do the help pop-up immediately and then not send the chat message to the server at all.

Emperorlou: Don't you think?

Emperorlou: If you still want to do it this way I'll have to add it in my end. It's OK whatever you decide you want to do.

tacobowl8:
Put /help to only trigger clientside and not put anything into the chat.  Also added /trade <character name>

Initium-Odp/war/odp/javascript/messager-impl.js
```diff
@@ -57,6 +57,12 @@ messager.onNotificationMessage = function(message)
 
 messager.onChatMessage = function(chatMessage)
 {
+	if (chatMessage.message.toLowerCase().startsWith("/")) {
```
Emperorlou: To lower case unnecessary here. Also......... Are you sure Js has that function for strings?

Initium-Odp/war/odp/javascript/messager-impl.js
```diff
@@ -145,6 +151,18 @@ messager.onChatMessage = function(chatMessage)
 	notifyNewMessage(chatMessage.code);
 };
 
+function tryChatCommands(chatMessage)
+{
+	if (chatMessage.message == "/help") {
```
Emperorlou: Here is where you would want to call to lower case

PR #470, opened 2017-01-27
Creating mapPlow command

Dias-Flac:


Initium-Odp/war/odp/javascript/Sandbox.js
```diff
@@ -1129,6 +1095,160 @@ function CoordObject(xGridCoord, yGridCoord) {
     this.yGridCoord = yGridCoord;
 }
 
-function mapPlow(coord) {
-    
+function mapPlow() {
+    if (cursorObject != "") {
+        doCommand(event, "MapPlow", {"xGridCoord": cursorObject.xGridCoord, "yGridCoord": cursorObject.yGridCoord}, function (data, error) {
+            if (error) return;
+            updateGridFromServer(data);
```
Emperorlou: We should call this update2DViewport() since it would do any and all updates relating to the viewport. Just a context/end-dev thing.

PR #480, opened 2017-01-30
Adding updating map to script.js call

Dias-Flac:


Initium-Odp/war/odp/javascript/script.js
```diff
@@ -2263,6 +2263,10 @@ function doCommand(eventObject, commandName, parameters, callback)
 		// Here we display the system message if there was a system message
 		if (data.message!=null && data.message.length>0)
 			popupMessage("System Message", data.message);
+		
+		if (data._2dViewportUpdates!=null && data._2dViewportUpdates.length>0) {
```
Emperorlou: It's not an array is it? I think just the initial null check is probably sufficient I suppose..

PR #504, opened 2017-02-22
Added NullChecks to item Titles

dbenson24:


Initium-Odp/src/com/universeprojects/miniup/server/GameUtils.java
```diff
@@ -851,20 +851,49 @@ else if (finalQuality<3)
...
+    		if (bc == null)
+    			bc = 0;
+    			
+    		Long dr = (Long)item.getProperty("damageReduction");
+    		if (dr == null)
+    			dr = 0;
+    			
+    		String bludge = (String)item.getProperty("blockBludgeoningCapability");
+    		if (bludge == null)
+    			bludge = "N";
```
Emperorlou: When a capability == null, it is assumed to be Average

Initium-Odp/src/com/universeprojects/miniup/server/GameUtils.java
```diff
@@ -851,20 +851,49 @@ else if (finalQuality<3)
...
+    		if (dp == null) 
+    			dp = 0;
+    		
+    		Long bc = (Long)item.getProperty("blockChance");
+    		if (bc == null)
+    			bc = 0;
+    			
+    		Long dr = (Long)item.getProperty("damageReduction");
+    		if (dr == null)
+    			dr = 0;
```
Emperorlou: When DR is null, it is assumed to be 10

PR #507, opened 2017-02-23
Moved item titles over to minitips, added mouse handlers for creating…

dbenson24:
… and removing minitips

### DIFF
--- a/Initium-Odp/war/odp/javascript/script.js
+++ b/Initium-Odp/war/odp/javascript/script.js
@@ -90,6 +90,16 @@ $(window).ready(function(e){
 		$(event.currentTarget).parent().find("input:checkbox").click();
 	});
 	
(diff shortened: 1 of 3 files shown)


Initium-Odp/war/odp/javascript/script.js
```diff
@@ -90,6 +90,16 @@ $(window).ready(function(e){
 		$(event.currentTarget).parent().find("input:checkbox").click();
 	});
 	
+	// Handlers for the minitip overlays
+	$('#page-popup-root').on("mouseover", "[minitip]", function(event) {
+		$(this).append('<div class="minitip">' + $(this).attr("title") + '</div>');
```
Emperorlou: Shouldn't be limited to page pop-up root since I could use this higher up

PR #512, opened 2017-02-23
added <br/> to minitip, fixed double precision

dbenson24:


Initium-Odp/src/com/universeprojects/miniup/server/GameUtils.java
```diff
@@ -872,8 +872,8 @@ private static String getItemMiniTip(CachedEntity item) {
 	    		if (critChance == null)
 	    			critChance = new Long(0);
 	    			
-	    		sb.append(diceRoll + "x" + critMultiplier + " " + critChance + "% ");
-	    		sb.append("(" + getWeaponMaxDamage(item) + "/" + getWeaponAverageDamage(item) + ") ");
+	    		sb.append(diceRoll + "x" + String.format("%1$.1f", critMultiplier) + " " + critChance + "% ");
```
Emperorlou: Actually here you could use GameUtils.formatNumber()

PR #514, opened 2017-02-26
Updates chat help with /exchange

tacobowl8:


Initium-Odp/war/odp/chatHelp.html
```diff
@@ -24,6 +24,9 @@
 					first though! <a onclick="viewManageStore()">You can do that
 						here</a>
 				</li>
+				<li>/exchange - This shows a link to the Global Premium Membership 
+					Token Exchange page. <a onclick="viewExchange()">
```
Emperorlou: the <a> is missing a </a> plus some text inside. ^^

PR #519, opened 2017-03-02
Sorting inventories when they load

dbenson24:


Initium-Odp/war/odp/javascript/script.js
```diff
@@ -874,7 +874,81 @@ function loadInventoryAndEquipment()
 
 function loadInventory()
 {
-	$("#inventory").load("/odp/inventorylist.jsp?ajax=true");
+	$("#inventory").load("/odp/inventorylist.jsp?ajax=true", function(){
```
Emperorlou: Could we put all this logic into a function so we can apply it wherever items are listed in an organized manner?

PR #527, opened 2017-03-05
Only group owners can kick group administrators.

armorproof:


Initium-Odp/src/com/universeprojects/miniup/server/commands/CommandGroupMemberKick.java
```diff
@@ -77,6 +77,14 @@ public final void run(final Map<String, String> parameters)
 		{
 			throw new UserErrorMessage("The creator cannot be kicked you fool.");
 		}
+		
+		// If the target is an admin and the current player is not the group creator,
+		if (kickCharacter.getProperty("groupStatus").equals("Admin") &&
+				!(GameUtils.equals((Key)admin.getKey(),
```
Emperorlou: No need to cast to key here

Initium-Odp/src/com/universeprojects/miniup/server/commands/CommandGroupMemberKick.java
```diff
@@ -77,6 +77,14 @@ public final void run(final Map<String, String> parameters)
...
+						(Key)group.getProperty("creatorKey"))))
+		{
+			throw new UserErrorMessage("You cannot kick a group member of the same rank.");
```
Emperorlou: Since there is no mechanic for rank, better to use different wording so to not confuse the noobs.

armorproof:
I got the casting from the code earlier in the class
```
if (GameUtils.equals((Key) group.getProperty("creatorKey"), (Key) kickCharacter.getKey()))
```
Is it unnecessary here too?

PR #544, opened 2017-03-26
[F]irst time using Java, be gentle ;)

papamarsh:
Autofix command for characters stuck in a location with no discoverable paths out.

### DIFF
--- a/Initium-Odp/src/com/universeprojects/miniup/server/commands/CommandCharacterAutofixStuckInLocation.java
+++ b/Initium-Odp/src/com/universeprojects/miniup/server/commands/CommandCharacterAutofixStuckInLocation.java
@@ -1,0 +1,68 @@
+package com.universeprojects.miniup.server.commands;
+
+import java.util.Map;
+import java.util.List;
+
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+
+import com.google.appengine.api.datastore.Key;
+import com.universeprojects.cacheddatastore.CachedDatastoreService;
+import com.universeprojects.cacheddatastore.CachedEntity;
+import com.universeprojects.miniup.CommonChecks;
+import com.universeprojects.miniup.server.GameUtils;
+import com.universeprojects.miniup.server.ODPDBAccess;
+import com.universeprojects.miniup.server.WebUtils;
+import com.universeprojects.miniup.server.commands.framework.Command;
+import com.universeprojects.miniup.server.commands.framework.UserErrorMessage;
+
+/**
+ * Checks to see if a character is stuck in a location with no discoverable paths available. If the character
+ * is stuck, their locationKey is set to the parentLocationKey of their current location. 
+ * 
+ * @author papamarsh
+ * 
+ */
+public class CommandCharacterAutofixStuckInLocation extends Command {
+
+	public CommandCharacterAutofixStuckInLocation(ODPDBAccess db, HttpServletRequest request,
+			HttpServletResponse response) {
+		super(db, request, response);
+	}
+
+	@Override
+	public void run(Map<String, String> parameters) throws UserErrorMessage {
+		ODPDBAccess db = ODPDBAccess.getInstance(request);
+		CachedDatastoreService ds = getDS();
+		
+		//grab the current character, its location, and the location's parent
+		CachedEntity character = db.getCurrentCharacter();
+		CachedEntity currentLocation = db.getEntity((Key) character.getProperty("locationKey"));
+		CachedEntity parentLocation = db.getParentLocation(ds, currentLocation);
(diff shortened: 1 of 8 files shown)


papamarsh:
Not sure where the command trigger should go.. The settings page doesn't really make sense and the profile page is pretty cluttered with random stuff.

papamarsh:
Added another commit for creating the autofix page popup and calling the command (Not very confident that it's all correct though, will definitely need checking over)

Initium-Odp/src/com/universeprojects/miniup/server/commands/CommandCharacterAutofixStuckInLocation.java
```diff
@@ -0,0 +1,70 @@
...
+
+	@Override
+	public void run(Map<String, String> parameters) throws UserErrorMessage {
+		ODPDBAccess db = ODPDBAccess.getInstance(request);
+		CachedDatastoreService ds = getDS();
+		
+		//grab the current character, its location, and the location's parent
+		CachedEntity character = db.getCurrentCharacter();
+		CachedEntity currentLocation = db.getEntity((Key) character.getProperty("locationKey"));
+		CachedEntity parentLocation = db.getParentLocation(ds, currentLocation);
```
Emperorlou: The currentLocation could be null (in rare case, but given the nature of the command we should probably handle that). If the location is null, we should assume it will be set on the next refresh. Perhaps try exiting early and calling setJavascriptResponse(JavascriptResponse.FullPageRefresh) to force the location to be set first.

Initium-Odp/src/com/universeprojects/miniup/server/commands/CommandCharacterAutofixStuckInLocation.java
```diff
@@ -0,0 +1,70 @@
...
+		{
+			throw new UserErrorMessage("Nice Try! But there's no escape from Hell :)");
+		}
+		
+		//list the paths connected to the character's location
+		List<CachedEntity> pathList = db.getPathsByLocation(currentLocation.getKey());
+		for(CachedEntity checkPath:pathList)
+		{
+			CachedEntity checkPathLocation1 = db.getEntity((Key) checkPath.getProperty("location1Key"));
+			CachedEntity checkPathLocation2 = db.getEntity((Key) checkPath.getProperty("location1Key"));
```
Emperorlou: you meant location2Key here

Emperorlou: Since this entire command is supposed to fail gracefully, maybe we should assume the locations could be null as well and skip a path if it's malformed..... actually, the easiest thing could be to wrap everything inside the for loop in a try-catch and just ignore any exceptions that do come up.

papamarsh: I'm not sure I'm following what you mean here. If either locationKey of a path is null, then that path wouldn't be a valid exit route, so it should be ignored by the IF statement inside the FOR loop. 

The way it's set up right now (assuming it works correctly) is to just check every path that has the player's current location in one of its locationKeys. Ignore any that can't be used (ie. one locationKey is null or the discoveryChance is <0). If the loop sees a valid path that is discoverable, then throw the popup that a valid path exists.

Initium-Odp/war/odp/ajax_autofix.jsp
```diff
@@ -0,0 +1,30 @@
+<%@ taglib uri='http://java.sun.com/jsp/jstl/core' prefix='c'%>
```
Emperorlou: This is the older way of adding a page. Now pages go in war/web-inf/odppages and must have an accompanying controller to go with it. In your case the controller will be basically empty, but it's a good chance to try your hand at it.

Initium-Odp/src/com/universeprojects/miniup/server/commands/CommandCharacterAutofixStuckInLocation.java
Emperorlou: Btw you still need to fix the property name here.

As for the other thing... I see what you mean. I was just saying that we should wrap everything inside of the loop in a try-catch so that if anything is null in there (the only possibility at this point is that the path is null) then it would at least keep trying to find a valid path. I think we'll want to do more in this loop too. For example, if the character is in a combatsite, he shouldn't have to explore to find a path, it should be there already.  In that case, having the try-catch will be even more handy since the location could be null. It just makes it so you don't have to do null checks for everything. It's not the usual scenario of course, normally you don't want to swallow exceptions like that, but this very specific scenario is about trying to provide a workaround for unknown or known bugs, so it works.

PR #548, opened 2017-04-01
Autofix command for the bug where unconscious mode isn't set (a few other random tweaks too)

papamarsh:


Initium-Odp/src/com/universeprojects/miniup/server/commands/CommandAutofixDeathModeNotSet.java
```diff
@@ -0,0 +1,87 @@
...
+		//grab the current character and its location
+		CachedEntity character = db.getCurrentCharacter();
+		CachedEntity currentLocation = db.getEntity((Key) character.getProperty("locationKey"));
+		
+		//grab the character's current hitpoints and mode
+		Double hitpoints = (Double) character.getProperty("hitpoints");
+		String mode = (String) character.getProperty("mode");
+		
+		//check for non-positive HP while mode is not either unconscious or dead
+		if(hitpoints<=0 && mode!="UNCONSCIOUS" && mode!="DEAD")
```
Emperorlou: Strings cannot be compared this way in java. use .equals(). mode can also be null, so use "UNCONSCIOUS".equals(mode) to avoid additional null checks.

PR #603, opened 2017-06-09
Remove party from combat after character is killed

jenga201:
In addition to removing attacking character from combat,
also remove party members if needed.

### DIFF
--- a/Initium-Odp/src/com/universeprojects/miniup/server/ODPDBAccess.java
+++ b/Initium-Odp/src/com/universeprojects/miniup/server/ODPDBAccess.java
@@ -4294,6 +4294,14 @@ public class ODPDBAccess
 					attackingCharacter.setProperty("combatType", null);
 					attackingCharacter.setProperty("locationEntryDatetime", new Date());
 	
+					// Make the attacking character party no longer in combat mode
+					List<CachedEntity> party = getParty(ds, attackingCharacterFinal);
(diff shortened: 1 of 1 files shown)


Initium-Odp/src/com/universeprojects/miniup/server/ODPDBAccess.java
```diff
@@ -4294,6 +4294,14 @@ public String doCharacterKilled(CachedEntity user, final CachedEntity characterT
 					attackingCharacter.setProperty("combatType", null);
 					attackingCharacter.setProperty("locationEntryDatetime", new Date());
 	
+					// Make the attacking character party no longer in combat mode
+					List<CachedEntity> party = getParty(ds, attackingCharacterFinal);
+					if (party != null) {
+						setPartiedField(party, attackingCharacter, "mode", ODPDBAccess.CHARACTER_MODE_NORMAL);
+						setPartiedField(party, attackingCharacter, "combatant", null);
+						setPartiedField(party, attackingCharacter, "combatType", null);
+						setPartiedField(party, attackingCharacter, "locationEntryDatetime", new Date());
```
Emperorlou: Check that setPartiedField() actually calls ds.put() on each of the party members too (my guess is it doesn't).  I think there is a method for putting all party members. Look for it where setPartiedField() was written, it would be near there.

PR #605, opened 2017-06-09
Updating window.verifyCode on main page update

jenga201:
Setting to empty string if user is not logged in


jenga201:
updateVerifyCode now only called on character switch

PR #610, opened 2017-06-10
renderAchievements

durpcity:
added renderAchievements method within GameUtils
created Achievement domain


Initium-Odp/src/com/universeprojects/miniup/server/domain/Achievement.java
```diff
@@ -0,0 +1,50 @@
+package com.universeprojects.miniup.server.domain;
```
Emperorlou: Domain stuff isn't used and it's unclear if it will ever be at this point. Fyi

Initium-Odp/src/com/universeprojects/miniup/server/ODPDBAccess.java
```diff
@@ -6263,4 +6263,9 @@ public boolean hasNewGameMessages()
 		// TODO Auto-generated method stub
 		return false;
 	}
+
+	public List<CachedEntity> getAchievementsFor(Key key) {
```
Emperorlou: What is expected here? The argument is just 'key' so I'm not sure what the intention is here

Emperorlou: Oh I see its for a characterKey

PR #691, opened 2017-08-11
Changes to item movement

coreyalexbarker:
Files modified (1): ODPDBAccess.java

Note: Partial stack movement messaging could be cleaner, but will probably have to wait until a command is created for it.

### DIFF
--- a/Initium-Odp/src/com/universeprojects/miniup/server/ODPDBAccess.java
+++ b/Initium-Odp/src/com/universeprojects/miniup/server/ODPDBAccess.java
@@ -2220,7 +2220,7 @@ public class ODPDBAccess
 	public boolean tryMoveItem(CachedEntity character, CachedEntity item, CachedEntity newContainer, Long requestQuantity) throws UserErrorMessage
 	{
 		CachedDatastoreService ds = getDB();
-		
+
 		if (CommonChecks.checkItemIsMovable(item)==false)
 			throw new UserErrorMessage("You cannot move this item.");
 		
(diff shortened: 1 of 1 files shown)


Initium-Odp/src/com/universeprojects/miniup/server/ODPDBAccess.java
```diff
@@ -2320,24 +2357,30 @@ else if (startKind.equals("Location"))
 					throw new UserErrorMessage("You are not near this item, you cannot pick it up.");
 				
 				// Check if the character can actually carry something else or if its all too heavy...
-				Long itemWeight = getItemWeight(item);
+				// And if quantity is specified, measure for specified quantity rather than whole stack
+				Long itemWeight = (requestQuantity != null) ? requestQuantity*itemSingleWeight : getItemWeight(item);
+
+				// Again, not sure if this check is necessary, but leaving it for now
```
Emperorlou: I think it is necessary because getItemWeight() will probably return null if the weight of the item's weight is null, which can happen legitimately.

Initium-Odp/src/com/universeprojects/miniup/server/ODPDBAccess.java
```diff
@@ -2359,44 +2402,110 @@ else if (startKind.equals("Item"))
...
+			final Long oldQuantity = itemQuantity;
+			final Long newQuantity = requestQuantity;
+
+			try {
+				item = (CachedEntity) new Transaction<CachedEntity>(ds) {
+
+					@Override
+					public CachedEntity doTransaction(CachedDatastoreService ds) throws AbortTransactionException {
+
+						CachedEntity item = getEntity(itemKey);
```
Emperorlou: I think there is a refetch() method which you might want to use instead of getEntity(). It's specifically made for refetching entities that were originally fetched outside of a transaction for use inside of a transaction.

Initium-Odp/src/com/universeprojects/miniup/server/ODPDBAccess.java
```diff
@@ -2359,44 +2402,110 @@ else if (startKind.equals("Item"))
...
+
+						ds.preallocateIdsFor(itemKind, 1);
```
Emperorlou: You don't need to preallocate here. as the call to getPreallocatedIdFor() will automatically preallocate 1 ID when called if none have been preallocated already.

That being said, you don't need to preallocate at all unless you need to save the key from this particular entity (your newItem) in a field on another item. You can just let the DB generate the ID when the transaction is complete instead (saves a round trip to the DB).

PR #708, opened 2017-09-08
Null Item

durpcity:
Null amount for Sale adds Museum to SaleItem status, which cannot be purchased.


Emperorlou:
Hows it goin?

Emperorlou:
Closed due to age and the fact that the implementation was a bit incomplete.

PR #743, opened 2017-12-01
Museum Items

durpcity:


Initium-Odp/war/odp/ajax_viewstore.jsp
```diff
@@ -111,6 +111,10 @@
         }
         
         // These are only used for the sorting method db.sortSaleItemList()
+        if ("Museum".equals(saleItem.getProperty("status"))) {
+        	item.setProperty("dogecoins",0);
```
Emperorlou: This doesn't look right I think?

Emperorlou:
Closed due to age and the fact that the item was never saved to the DB after updating it.

PR #755, opened 2018-01-15
Exchange controller fix

Eriostafel:
added lines 115-118
edited line 120

### DIFF
--- a/Initium-Odp/src/com/universeprojects/miniup/server/controllers/ExchangeController.java
+++ b/Initium-Odp/src/com/universeprojects/miniup/server/controllers/ExchangeController.java
@@ -112,7 +112,12 @@ public class ExchangeController extends PageController {
 	        
 	        
 	        // These are only used for the sorting method db.sortSaleItemList()
-	        item.setProperty("store-dogecoins", saleItem.getProperty("dogecoins"));
+		Double storeSale = (Double)sellingCharacter.getProperty("storeSale");
+		if(storeSale==null) storeSale = Double.valueOf(100);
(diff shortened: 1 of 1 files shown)


Initium-Odp/src/com/universeprojects/miniup/server/controllers/ExchangeController.java
```diff
@@ -112,7 +112,12 @@ protected final String processRequest(HttpServletRequest request, HttpServletRes
 	        
 	        
 	        // These are only used for the sorting method db.sortSaleItemList()
-	        item.setProperty("store-dogecoins", saleItem.getProperty("dogecoins"));
+		Double storeSale = (Double)sellingCharacter.getProperty("storeSale");
+		if(storeSale==null) storeSale = Double.valueOf(100);
```
Emperorlou: you can just say 100 or 100d instead of doing the valueOf thing

Initium-Odp/src/com/universeprojects/miniup/server/controllers/ExchangeController.java
```diff
@@ -112,7 +112,12 @@ protected final String processRequest(HttpServletRequest request, HttpServletRes
...
+		Long coins = (Long)saleItem.getProperty("dogecoins");
+		coins = Long.valueOf(Math.round(coins.doubleValue()*((storeSale.doubleValue())/100d)));    
+		    
+	        item.setProperty("store-dogecoins", (Object)coins);
```
Emperorlou: You don't have to cast to Object because ALL objects are Objects already.

PR #762, opened 2018-01-18
viewcharactermini.jsp made into separate java and jsp files. 

niiloc:
rel's fixed to mirror the changes.

### DIFF
--- a/Initium-Odp/src/com/universeprojects/miniup/server/controllers/ViewCharacterMiniController.java
+++ b/Initium-Odp/src/com/universeprojects/miniup/server/controllers/ViewCharacterMiniController.java
@@ -2,20 +2,22 @@ package com.universeprojects.miniup.server.controllers;
 
 import java.io.IOException;
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.Comparator;
 import java.util.List;
-import java.util.Collections;
 
 import javax.servlet.ServletException;
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
 
 import com.google.appengine.api.datastore.Key;
+import com.google.appengine.api.datastore.KeyFactory;
 import com.universeprojects.cacheddatastore.CachedEntity;
 import com.universeprojects.cacheddatastore.ShardedCounterService;
 import com.universeprojects.miniup.CommonChecks;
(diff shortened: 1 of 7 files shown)


Initium-Odp/src/com/universeprojects/miniup/server/controllers/ViewCharacterMiniController.java
```diff
@@ -0,0 +1,219 @@
...
+		super("viewcharactermini");
+	}
+	
+	@Override
+	protected String processRequest(HttpServletRequest request,
+			HttpServletResponse response) throws ServletException, IOException {
+		response.setHeader("Access-Control-Allow-Origin", "*");     // This is absolutely necessary for phonegap to work
+		
+		ODPDBAccess db = ODPDBAccess.getInstance(request);
+		CachedEntity character = db.getCurrentCharacter();
```
SPFiredrake: `character` should be the character we're trying to view. Look in `ViewItemController` to see how we get the `characterId` attribute from the request to get a character entity.

```Long charId = WebUtils.getLongParam(request, "characterId");
if (charId==null) return null;
Key characterKey = KeyFactory.createKey("Character", charId);
CachedEntity character = db.getEntity(characterKey); ```

Don't forget to validate that the character exists after retrieving.

File: Initium-Odp/src/com/universeprojects/miniup/server/controllers/ViewCharacterMiniController.java [hunk 2]
```diff
@@ -0,0 +1,219 @@
...
+	String type = (String)character.getProperty("type");
+	if (type==null || type.equals(""))
+		type = "PC";
+	request.setAttribute("type", type);		
+
+	// If the character we're trying to inspect is an NPC and he is still alive, don't allow the inspection
+	if (type.equals("NPC") && (Double)character.getProperty("hitpoints")>0d)
+		throw new RuntimeException("Should not have been able to view this character.");
+	
+	boolean isSelf = GameUtils.equals(db.getCurrentUser().getKey(), character.getProperty("userKey"));
```
SPFiredrake: We've already retrieved the current user at the top of this method (`user` variable). Shortcut to `user.getKey()` instead of `db.getCurrentUser().getKey()`.

PR #867, opened 2018-07-24
Attempted to add /settings chat command.

ThatGuyGW:


Initium-Odp/src/com/universeprojects/miniup/server/controllers/SettingsController.java
```diff
@@ -0,0 +1,42 @@
+package com.universeprojects.miniup.server.controllers;
```
Emperorlou: Why was this file added? Was it not there already?

Initium-Odp/src/com/universeprojects/miniup/server/ServletEventServer.java
```diff
@@ -188,7 +188,12 @@ else if (accountKey.getKind().equals("User"))
 				{
 					message = "<a onclick='viewStore("+character.getKey().getId()+")'>[Visit Store]</a> "+message.substring(9);
 				}
-				
+
+				if (message.startsWith("/settings"))
+				{
+                    message = "<a onclick='viewSettings()'>[Tweak your Settings]</a> "+message.substring(8);
```
Emperorlou: I think the substring(8) actually needs to be substring(9)

Initium-Odp/war/odp/chatHelp.html
```diff
@@ -24,6 +24,10 @@
 					customization page. <a onclick="customizeItemOrderPage()">You
 						can also find it here</a>
 				</li>
+				<li>/merchant - This shows a link to your settings page where
```
Emperorlou: copypasta - looks like /merchant should be /settings

PR #962, opened 2020-08-04
Fix HCM Leaderboard; characters with a score of 0 shouldn't appear.

Evanosity:
added a new check, if a character would be loaded onto the leaderboard but has an empty score, it gets cut. This should also keep non-HCM characters from appearing on the leaderboard.

### DIFF
--- a/Initium-Odp/src/com/universeprojects/miniup/server/controllers/HCMLeaderboardController.java
+++ b/Initium-Odp/src/com/universeprojects/miniup/server/controllers/HCMLeaderboardController.java
@@ -51,8 +51,8 @@ public class HCMLeaderboardController extends PageController
 		for(int i = 0; i<characters.size(); i++)
 		{
 			CachedEntity character = characters.get(i);
-						
-			if((int)character.getProperty("hardcoreRank")<=0||character.getProperty("hardcoreRank")==null) continue;
+			
+			if(character.getProperty("hardcoreRank")==null || (int)character.getProperty("hardcoreRank")<=0) continue;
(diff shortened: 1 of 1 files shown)


Initium-Odp/src/com/universeprojects/miniup/server/controllers/HCMLeaderboardController.java
```diff
@@ -52,6 +52,8 @@ protected String processRequest(HttpServletRequest request, HttpServletResponse
 		{
 			CachedEntity character = characters.get(i);
 			
+			if((int)character.getProperty("hardcoreRank")<=0) continue;
```
Emperorlou, line 55: This can be null here, you'd need to do a null check first.

Initium-Odp/src/com/universeprojects/miniup/server/controllers/HCMLeaderboardController.java
```diff
@@ -51,6 +51,8 @@ protected String processRequest(HttpServletRequest request, HttpServletResponse
 		for(int i = 0; i<characters.size(); i++)
 		{
 			CachedEntity character = characters.get(i);
+						
+			if((int)character.getProperty("hardcoreRank")<=0||character.getProperty("hardcoreRank")==null) continue;
```
Emperorlou, line 55: it does the checks in left to right order, so it would never check the null before crashing on the (int) cast.

Also, you should add some spaces in there specifically around the || and optionally around the <= and ==

PR #963, opened 2020-08-12
Various fixes + QoL changes

Evanosity:
Allows an item to have multiple equip "options", each with multiple slots. For example; an item slot of "Chest and Legs, Chest and Shirt and Legs" would now work properly(not that you would want to do that....).

### DIFF
--- a/Initium-Odp/src/com/universeprojects/miniup/CommonChecks.java
+++ b/Initium-Odp/src/com/universeprojects/miniup/CommonChecks.java
@@ -65,6 +65,16 @@ public abstract class CommonChecks
 	}
 	
 	/**
+	 * Checks if a character is zombified.
+	 * @param character
+	 * @return
+	 */
+	public static boolean checkCharacterIsZombie(CachedEntity character) {
+		if("Zombie".equals(character.getProperty("status"))) return true;
+		return false;
+	}
+	
+	/**
 	 * Checks if a character is unconscious. 
 	 * 
 	 * This will return false if the character is actually dead.
@@ -448,7 +458,7 @@ public abstract class CommonChecks
 
 	public static boolean checkCharacterIsIncapacitated(CachedEntity character)
 	{
-		if (checkCharacterIsDead(character) || checkCharacterIsUnconscious(character))
+		if (checkCharacterIsDead(character) || checkCharacterIsUnconscious(character) || checkCharacterIsZombie(character))
 			return true;
 		return false;
 	}
--- a/Initium-Odp/src/com/universeprojects/miniup/CommonChecks.java
+++ b/Initium-Odp/src/com/universeprojects/miniup/CommonChecks.java
@@ -58,8 +58,9 @@ public abstract class CommonChecks
 	public static boolean checkCharacterIsDead(CachedEntity character)
 	{
 		Double hitpoints = (Double)character.getProperty("hitpoints");
-		if (hitpoints<=0 && "DEAD".equals(character.getProperty("mode")))
+		if (hitpoints<=0 && ("DEAD".equals(character.getProperty("mode")) || "RESPANWED".equals(character.getProperty("mode")))) {
 			return true;
+		}
 		
 		return false;
 	}
@@ -89,6 +90,16 @@ public abstract class CommonChecks
 			return true;
 		
 		return false;
+	}
+	
+	/**
+	 * Checks if a character is respawnable.
+	 * @param character
+	 * @return
+	 */
+	public static boolean checkCharacterIsRespawnable(CachedEntity character) {
+		if("RESPAWNED".equals(character.getProperty("mode"))) return false;
+		return true;
 	}
 	
 	/**
--- a/Initium-Odp/src/com/universeprojects/miniup/CommonChecks.java
+++ b/Initium-Odp/src/com/universeprojects/miniup/CommonChecks.java
@@ -71,8 +71,7 @@ public abstract class CommonChecks
 	 * @return
 	 */
 	public static boolean checkCharacterIsZombie(CachedEntity character) {
-		if("Zombie".equals(character.getProperty("status"))) return true;
-		return false;
+		return "Zombie".equals(character.getProperty("status"));
 	}
 	
 	/**
--- a/Initium-Odp/src/com/universeprojects/miniup/CommonChecks.java
+++ b/Initium-Odp/src/com/universeprojects/miniup/CommonChecks.java
@@ -58,7 +58,7 @@ public abstract class CommonChecks
 	public static boolean checkCharacterIsDead(CachedEntity character)
 	{
 		Double hitpoints = (Double)character.getProperty("hitpoints");
-		if (hitpoints<=0 && ("DEAD".equals(character.getProperty("mode")) || "RESPANWED".equals(character.getProperty("mode")))) {
+		if (hitpoints <= 0 && ("DEAD".equals(character.getProperty("mode")))) {
 			return true;
 		}
 		
@@ -89,16 +89,6 @@ public abstract class CommonChecks
 			return true;
 		
 		return false;
-	}
-	
-	/**
-	 * Checks if a character is respawnable.
-	 * @param character
-	 * @return
-	 */
-	public static boolean checkCharacterIsRespawnable(CachedEntity character) {
-		if("RESPAWNED".equals(character.getProperty("mode"))) return false;
-		return true;
 	}
 	
 	/**
--- a/Initium-Odp/src/com/universeprojects/miniup/server/ODPDBAccess.java
+++ b/Initium-Odp/src/com/universeprojects/miniup/server/ODPDBAccess.java
@@ -784,6 +784,34 @@ public class ODPDBAccess
 		}
 		return null;
 	}
+	
+	/**
+	 * This returns an alphabetically sorted list of all characters associated with a userkey.
+	 * This will also filter out all invalid characters, notable dead and zombie.
+	 * @param userKey
+	 * @return
+	 */
+	public List<CachedEntity> getAlphabetSortedValidCharactersByUser(Key userKey){
+		List<CachedEntity> characters = query.getFilteredList("Character", "userKey", userKey);
+		Collections.sort(characters, new Comparator<CachedEntity>()
+		{
+			@Override
+			public int compare(CachedEntity o1, CachedEntity o2)
+			{
+				return ((String)o1.getProperty("name")).compareTo((String)o2.getProperty("name"));
+			}
+		});
+		
+		//this will validate all the characters. Filter out dead and zambie.
+		List<CachedEntity> toReturn = new ArrayList<>();
+		for(CachedEntity c:characters) {
+			if (c.getProperty("name").toString().startsWith("Dead ") == false && "Zombie".equals(c.getProperty("status")) == false) {
+				toReturn.add(c);
+			}
+		}
+		
+		return toReturn;
+	}
 
 	public CachedEntity getUserByEmail(String email)
 	{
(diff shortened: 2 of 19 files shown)


Evanosity:
The fix for duping alts will also have to reflected in the schema.

Initium-Odp/src/com/universeprojects/miniup/CommonChecks.java
```diff
@@ -58,12 +58,23 @@ public static boolean checkCharacterIsBusy(CachedEntity character)
...
 	
+	/**
+	 * Checks if a character is zombified.
+	 * @param character
+	 * @return
+	 */
+	public static boolean checkCharacterIsZombie(CachedEntity character) {
+		if("Zombie".equals(character.getProperty("status"))) return true;
+		return false;
+	}
```
SPFiredrake, line 76: Get rid of the `if`, and just return the comparison.

Initium-Odp/src/com/universeprojects/miniup/server/ODPDBAccess.java
```diff
@@ -84,7 +84,7 @@
 	
 	public enum CharacterMode
 	{
-		NORMAL, COMBAT, MERCHANT, TRADING, UNCONSCIOUS, DEAD
+		NORMAL, COMBAT, MERCHANT, TRADING, UNCONSCIOUS, DEAD, RESPANWED
```
SPFiredrake, line 87: Fix spelling of `RESPANWED`

Initium-Odp/src/com/universeprojects/miniup/server/ODPDBAccess.java
```diff
@@ -1572,21 +1600,56 @@ else if (equipSlotArr.length>1)
...
+						if(complexSlot[i].equals("2Hands")) {
+							complexSlot[i] = "LeftHand and RightHand";
+						}
+
+						
+						//if there is something equipped in one of the slots, we break.
+						if(character.getProperty("equipment" + complexSlot[i]) != null){
+							isValid = false;
+							break;
+						}
```
SPFiredrake, line 1634: In this scenario, `2Hands` slot will _never_ work in a complex slot, since you'll be checking for `equipmentLeftHand and RightHand` property on character entity.

Initium-Odp/src/com/universeprojects/miniup/server/scripting/jsaccessors/DBAccessor.java
```diff
@@ -322,6 +322,15 @@ public boolean unequipItem(EntityWrapper item, EntityWrapper currentCharacter)
...
 	
+	/**
+	 * Sends a game message to the specified characters
+	 */
+	public void sendMessageToCharacters(String message, Character... characters) {
+		for(Character target:characters) {
+			db.sendGameMessage(db.getDB(), target.wrappedEntity, message);
+		}
+	}
+
```
SPFiredrake, line 333: Can't do this. If an error occurs during the script, then these calls shouldn't happen. You need to queue up any game messages you want to send, and only send when the call was successful. Look at `ScriptEvent` for `sendGameUpdates` and `CommandScriptBase` for `!event.haltExecution` code path.

Initium-Odp/src/com/universeprojects/miniup/server/scripting/wrappers/EntityWrapper.java
```diff
@@ -201,6 +201,16 @@ public boolean setHardcoreMode(boolean hcmMode)
...
+	/**
+	 * JS numbers don't translate cleanly; here, we can make sure that we are actually passing a LONG.
+	 * Useful for changing DR, int mod, dex mod, str mod, etc.
+	 * @param property
+	 * @param value
+	 */
+	public void setLongProperty(String property, Long value) {
+		this.setProperty(property, value);
+	}
+
```
SPFiredrake, line 213: This isn't allowed. Certain fields we _don't_ allow changing through the wrapper objects. This would basically expose every single property to scripting environment, which we try to expose only through certain fields/methods.

Initium-Odp/war/WEB-INF/odppages/ajax_sublocations.jsp
```diff
@@ -1,21 +1,23 @@
...
-		<div class='sublocation-display' style='background-image:url(${location.bannerUrl})'></div>
-		<h5>${location.name}</h5>
-		<c:if test="${location.isForgettable=='true'}">
-			<div class='standard-button-highlight forget-button' onclick='doForgetCombatSite(event,${location.locationId})'>X</div>
-		</c:if>
-	</div>
-</c:forEach>
-<c:if test='${forgetAllCombatSitesHtml!=null }'>
-${forgetAllCombatSitesHtml}
+<c:if test='${locationsViewable==true }'>
```
SPFiredrake, line 4: I'm only seeing this attribute set to false when character is incapacitated, meaning it's never set to true. Make sure this is set to true in the default case of SublocationsController.

Initium-Odp/src/com/universeprojects/miniup/server/controllers/SublocationsController.java
```diff
@@ -44,12 +44,21 @@ protected final String processRequest(HttpServletRequest request, HttpServletRes
 		ODPDBAccess db = ODPDBAccess.getInstance(request);
 		try{InitiumPageController.requireLoggedIn(db);}catch(NotLoggedInException e){return InitiumPageController.loginMessagePage;}
 		
+		
+	    CachedEntity character = db.getCurrentCharacter();
+	    
+	    if(CommonChecks.checkCharacterIsIncapacitated(character)) {
+	    	request.setAttribute("locationsViewable", false);
```
SPFiredrake, line 51: Make sure attribute is set to true in the else case, otherwise sublocations will never be viewable.

Initium-Odp/war/odp/javascript/script.js
```diff
@@ -4916,21 +4854,57 @@ $(document).keyup(function(event){
...
+			window.history.replaceState({}, document.title, "/" + "main.jsp");		
+		else if (location.href.indexOf("/odp/experimental")>-1)
+			window.history.replaceState({}, document.title, "/" + "odp/experimental");		
+		else if (location.href.indexOf("/odp/full")>-1)
+			window.history.replaceState({}, document.title, "/" + "odp/full");		
+		
+		closeAllTooltips();
+		clearMakeIntoPopup();
+	});
+}
```
SPFiredrake, line 4903: No need to create a new method, since you can use optional params. If the original call doesn't specify a param value, then "undefined" is passed in (you can likely null coalesce to change undefined to null).

Initium-Odp/src/com/universeprojects/miniup/server/commands/CommandSwitchCharacter.java
```diff
@@ -38,15 +42,129 @@ public void run(Map<String, String> parameters) throws UserErrorMessage {
...
+			//This method will return all characters that we can switch to. Filters out zambies and dead characters right off the bat.
+			List<CachedEntity> characters = db.getAlphabetSortedValidCharactersByUser(user.getKey());
+			
+			switch(direction) {
+			
+			//UP in characterswitcher.
+			case 1:
+				for(int i = 0; i < characters.size(); i++) {
+					CachedEntity c = characters.get(i);
+					//here, we are searching for the currentcharacter, so we know its index within the sorted list.
```
SPFiredrake, line 65: Use case fall through to reduce code duplication.
```
case 1:
case 2:
    List<CachedEntity> characters = db.getAlphabetSortedValidCharactersByUser(user.getKey());
```

Initium-Odp/src/com/universeprojects/miniup/server/commands/CommandSwitchCharacter.java
```diff
@@ -38,15 +42,129 @@ public void run(Map<String, String> parameters) throws UserErrorMessage {
...
+					targetCharacterId = characters.get(i+1).getId();
+					break;
+				}
+				break;
+				
+			//Iterate through characters in the same party and userkey
+			case 3:
+				//here, we need to grab all the partied characters. then, filter by userkey.
+				List<CachedEntity> party = db.getParty(ds, currentCharacter);
+				if(party == null) throw new UserErrorMessage("You aren't in a party.");
```
SPFiredrake, line 100: Same thing. Case fall through. Retrieve party, then check `direction` to determine logic for retrieving character.

Initium-Odp/src/com/universeprojects/miniup/server/commands/CommandSwitchCharacter.java
```diff
@@ -63,6 +181,10 @@ public void run(Map<String, String> parameters) throws UserErrorMessage {
 		if ("Zombie".equals(targetCharacter.getProperty("status"))) {
 			throw new UserErrorMessage("You cannot switch to this character, it is now a zombie.");
 		}
+		
+		if ("RESPAWNED".equals(targetCharacter.getProperty("mode"))) {
```
SPFiredrake, line 185: CommonChecks.

Evanosity:
Changes on this branch:
- Fixes
- - Added common check for zombie, which means that zombies are no longer controllable.
- - When respawning, remove the userkey from the dead character. This fixes the issue of alive characters with a "Dead" prefix on their name being excluded from the character switcher. Also fixes the issue of people duping infinite alts.
- - Can no longer setup shop while dead; this means no more reviving characters.
- - Fix hasScript throwing a runtime exception if the given entity has no scripts.
- - Fixed paths being viewable while dead in experimental mode.
- Script functions
- - Added methods to set/get damage reduction, int mod, str mod and dex pen from items in script context.
- - Built functionality to send game messages to characters from script context.
- No more rerolling with items in inventory
- Built better character switching logic; can now specify a direction.
- Hotkey Changes
- - Removed L hotkey as it broke experimental mode. Players can just do ctrl+r anyways
- - Added U/J/K/L to switch characters.
- - Added N to open navigation
- - Added H to open the new hotkey page.
- - Added A to cancel all active long operations.
- Update map updated date to when it was last updated.

Initium-Odp/src/com/universeprojects/miniup/server/commands/CommandSwitchCharacter.java
```diff
@@ -54,61 +54,47 @@ public void run(Map<String, String> parameters) throws UserErrorMessage {
...
-					//if we're at the first element in the list
-					if(i == 0) {
-						targetCharacterId = characters.get(characters.size()).getId();
+					//Move UP in character list.
+					if(direction == 1) {
+						//if we're at the front of the list, loop back to the back.
+						if(i == 0) {
+							targetCharacterId = characters.get(characters.size()).getId();
+							break;
+						}
```
SPFiredrake, line 73: Will throw OOB exception, since you're trying to get the last array index + 1 (size() returns count, index is 0:size-1). 
Can simplify this processing step: `targetCharacterId = characters.get((i+ characters.size() - 1) % characters.size()).getId()`

Initium-Odp/src/com/universeprojects/miniup/server/commands/CommandSwitchCharacter.java
```diff
@@ -54,61 +54,47 @@ public void run(Map<String, String> parameters) throws UserErrorMessage {
...
+					//Move DOWN in character list.
+					if(direction == 2) {	
+						//if we're at the end of the list, loop back to the front.
+						if(i == characters.size()) {
+							targetCharacterId = characters.get(0).getId();
+							break;
+						}
+						
+						//otherwise, grab the next one.
+						targetCharacterId = characters.get(i+1).getId();
```
SPFiredrake, line 88: Can likely simplify both steps: `int dir = direction == 1 ? -1 : 1;`, then use that var with above suggested change instead of `- 1` to `+ dir`.

Initium-Odp/src/com/universeprojects/miniup/server/scripting/events/ScriptEvent.java
```diff
@@ -189,6 +187,35 @@ public boolean updatesGameState()
 		return jsResponse == JavascriptResponse.FullPageRefresh || reloadWidgets;
 	}
 	
+	private Map<Character, Set<String>> gameMessages = new HashMap<Character, Set<String>>();
```
SPFiredrake, line 190: Might want to change this to a `HashMap<Character, List<String>>` since order actually matters here. This would only prevent the exact same message from being sent, which is highly unlikely (and should be prevented in script anyway).

Initium-Odp/src/com/universeprojects/miniup/server/scripting/wrappers/Item.java
```diff
@@ -113,6 +113,32 @@ public boolean adjustDurability(Long addDura)
...
+	public void setStrMod(Long newStrMod) {
+		this.setProperty("strengthModifier", newStrMod);
+	}
+	public void setIntMod(Long newIntMod) {
+		this.setProperty("intelligenceModifier", newIntMod);
+	}
+	public void setDamageReduction(Long damageReduction) {
+		this.setProperty("damageReduction", damageReduction);
+	}
+
```
SPFiredrake, line 141: I'm not certain this is something we want to allow. Rogue dev can basically make OP items via scripting, with no log of the changes. Why we've always limited scripting to very few devs, since it has to be someone that can be trusted.

Initium-Odp/war/odp/javascript/script.js
```diff
@@ -4860,16 +4865,16 @@ $(document).keyup(function(event){
...
+		switchCharacter(event, null, 4);
 	}
 	else if (event.which==75){ // K - iterate through characters in same acc and party
-		switchCharacterByDirection(event, 3);
+		switchCharacter(event, null, 4);
 	}
 	else if (event.which==76){ // L - switch to party leader if possible
-		switchCharacterByDirection(event, 4);
+		switchCharacter(event, null, 4);
 	}
```
SPFiredrake: Fix direction parameter being sent to `switchCharacter` method.

Evanosity: I also modified the `switchCharacter` method to accept a direction, and to sanitize the value if no direction is sent

Initium-Odp/src/com/universeprojects/miniup/CommonChecks.java
```diff
@@ -448,7 +458,7 @@ public static boolean checkCharacterIsAdminOfHisGroup(CachedEntity character)
 
 	public static boolean checkCharacterIsIncapacitated(CachedEntity character)
 	{
-		if (checkCharacterIsDead(character) || checkCharacterIsUnconscious(character))
+		if (checkCharacterIsDead(character) || checkCharacterIsUnconscious(character) || checkCharacterIsZombie(character))
```
Emperorlou, line 461: Actually I'm not sure this is a good idea.... it might break zombies from being able to do things, like combat.

Emperorlou: Yea, a cursory check seems to indicate that this would immediately break zombies. I accepted a previous PR with this change so I have reverted this line in particular in master.

Emperorlou:
Closed this PR because I believe you said it is now split up into separate PRs.

PR #965, opened 2020-09-10
Slots

Evanosity:
-added a jsp page and the related controller for viewing items with the slottable aspect inside the character's inventory.
-filled out AspectSlotted to contain the command for adding a slottable item
-added a new commoncheck for checking if an aspect is present based on a string instead of a class; this allows us to check for aspects that don't have a class.

### DIFF
--- a/Initium-Odp/src/com/universeprojects/miniup/CommonChecks.java
+++ b/Initium-Odp/src/com/universeprojects/miniup/CommonChecks.java
@@ -589,6 +589,22 @@ public abstract class CommonChecks
 		
 		return false;
 	}
+	
+	public static boolean checkItemHasAspect(CachedEntity item, String aspectName) {
+		String raw = (String) item.getProperty("_aspect");
+		
+		//get rid of those pesky brackets.
+		raw = raw.substring(1, raw.length()-1);
+		
+		String[] aspects = raw.split(", ");
+		
+		for(String aspect:aspects) {
+			if(aspect.equals(aspectName)) {
+				return true;
+			}
+		}
+		return false;
+	}
 
 	public static boolean checkItemHasAspect(CachedEntity item, Class<? extends InitiumAspect> clazz)
 	{
@@ -609,6 +625,4 @@ public abstract class CommonChecks
 	{
 		return GameUtils.equals(location.getProperty("type"), "CityHall");
 	}
-
-
 }
--- a/Initium-Odp/src/com/universeprojects/miniup/server/aspects/AspectSlotted.java
+++ b/Initium-Odp/src/com/universeprojects/miniup/server/aspects/AspectSlotted.java
@@ -1,40 +1,197 @@
 package com.universeprojects.miniup.server.aspects;
 
+import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.List;
+import java.util.Map;
 
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+
+import com.google.appengine.api.datastore.EmbeddedEntity;
+import com.google.appengine.api.datastore.Key;
+import com.google.appengine.api.datastore.KeyFactory;
 import com.universeprojects.cacheddatastore.CachedEntity;
+import com.universeprojects.miniup.CommonChecks;
 import com.universeprojects.miniup.server.InitiumObject;
 import com.universeprojects.miniup.server.ItemAspect;
+import com.universeprojects.miniup.server.ODPDBAccess;
+import com.universeprojects.miniup.server.UserRequestIncompleteException;
+import com.universeprojects.miniup.server.ItemAspect.ItemPopupEntry;
+import com.universeprojects.miniup.server.commands.framework.Command;
+import com.universeprojects.miniup.server.commands.framework.UserErrorMessage;
 
-public class AspectSlotted extends ItemAspect
-{
+/**
+ * 
+ * @author Evan
+ *
+ */
+public class AspectSlotted extends ItemAspect {
 
-	public AspectSlotted(InitiumObject object)
-	{
+	public AspectSlotted(InitiumObject object) {
 		super(object);
 	}
 
+	//Here, we need to properly set the link. use aspectpet as a base.
 	@Override
 	public List<ItemPopupEntry> getItemPopupEntries(CachedEntity currentCharacter)
 	{
-		return null;
+		if (CommonChecks.checkItemIsAccessible(object.getEntity(), currentCharacter) && hasOpenSlots()) {
+			ItemPopupEntry ipe = new ItemPopupEntry("Insert an item into the slots", "Insert a slottable item into this item.",
+					"viewSlottableItems(event, " + object.getKey().getId() + ")");
+			return Arrays.asList(new ItemPopupEntry[]
+			{
+					ipe
+			});
+		}
+		else return null;
 	}
 
 	@Override
-	public String getPopupTag()
-	{
+	public String getPopupTag() {
 		if (getMaxCount()>0)
 			return "Slotted";
 		else
 			return null;
 	}
 
-	
-	public long getMaxCount()
-	{
-		Long count = (Long)this.getProperty("maxCount");
+	/**
+	 * Returns the number of max slots on this aspect.
+	 * @return
+	 */
+	public long getMaxCount() {
+		Long count = (Long)getProperty("maxCount");
 		if (count==null) count = 0L;
 		return count;
 	}
 	
+	/**
+	 * Returns true if this aspect has any open slots.
+	 * @return
+	 */
+	public boolean hasOpenSlots() {
+		return getMaxCount() < getOpenCount();
+	}
+	
+	/**
+	 * Returns the number of open slots on this aspect.
+	 * @return
+	 */
+	public long getOpenCount() {
+		Long max = (Long)getProperty("maxCount");
+		
+		List<EmbeddedEntity> currentlySlotted = (List<EmbeddedEntity>)getProperty("slotItems"));
+		
+		if(currentlySlotted == null) return max;
+		
+		return max-currentlySlotted.size();
+	}
+	
+	static {
+		addCommand("InsertItemToSlot", AspectSlotted.CommandInsertItemToSlot.class);
+	}
+	
+	
+	/**
+	 * given a slotted item and a slottable item, insert it into the slot and apply the
+	 * appropriate modifiers to the parent item.
+	 * @author Evan
+	 *
+	 */
+	public static class CommandInsertItemToSlot extends Command {
+
+		public CommandInsertItemToSlot(ODPDBAccess db, HttpServletRequest request, HttpServletResponse response) {
+			super(db, request, response);
+		}
+		
+		//Pass the ID of the item with slots and the item we're slotting in. We will need to sanitize it.
+		@Override
+		public void run(Map<String, String> parameters) throws UserErrorMessage, UserRequestIncompleteException{
+			
+			//generate the base item from the given ID, then do basic checks.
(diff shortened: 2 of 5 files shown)


Initium-Odp/src/com/universeprojects/miniup/server/aspects/AspectSlotted.java
```diff
@@ -1,40 +1,197 @@
...
+		public void run(Map<String, String> parameters) throws UserErrorMessage, UserRequestIncompleteException{
+			
+			//generate the base item from the given ID, then do basic checks.
+			Key baseKey = KeyFactory.createKey("Item", Long.parseLong(parameters.get("baseId")));
+			InitiumObject baseItem = db.getInitiumObject(baseKey);
+			
+			if(baseItem == null) throw new UserErrorMessage("Invalid Item ID.");
+			if(!CommonChecks.checkItemIsAccessible(baseItem.getEntity(), db.getCurrentCharacter()))
+				throw new UserErrorMessage("Nice try.");
+			if(!CommonChecks.checkItemHasAspect(baseItem.getEntity(), "Slotted"))
```
Emperorlou, line 118: This is unnecessary as there are checks on InitiumObject to know if it has a certain aspect or not.

Initium-Odp/src/com/universeprojects/miniup/server/aspects/AspectSlotted.java
```diff
@@ -1,40 +1,197 @@
...
+				throw new UserErrorMessage("This item doesn't have any slots.");
+			
+			//gets a compiler error in ODP environment but should work when deployed.
+			AspectSlotted slottedAspect = (AspectSlotted) baseItem.getInitiumAspect("Slotted");
```
Emperorlou, line 122: There shouldn't be compile errors in ODP, what is the error?

Initium-Odp/src/com/universeprojects/miniup/server/aspects/AspectSlotted.java
```diff
@@ -1,40 +1,197 @@
...
+			
+			//Write a method that copies a given entity to a new embeddedentity?
+			
+			EmbeddedEntity newEmbedded = null;
+			
+			List<EmbeddedEntity> currentlySlotted = (List<EmbeddedEntity>)baseItem.getProperty("Slotted:slotItems");
+			if(currentlySlotted == null) currentlySlotted = new ArrayList<>();
+			currentlySlotted.add(newEmbedded);
+			
+			List<Key> scriptsToAdd = (List<Key>) slottableItem.getProperty("Slottable:scripts");
```
Emperorlou, line 147: What if you're adding multiple of the same kind of item? You'll end up with multiple of the same script. Is that ok?

Emperorlou: We should perhaps instead grab the script keys from the slotted items, which will still be present in the embedded entity form of the item. No need to explicitly add it to the parent item. We would have to add some code to the place where we get the script keys though so that it also gets the ones on the slotted items.

Initium-Odp/src/com/universeprojects/miniup/server/aspects/AspectSlotted.java
Evanosity: I figured it out, the supertype of GameObject is missing from the ODP so I get some weird errors. I copied this code from the aspect pet so it will be fine.

Emperorlou: GameObject is an interface that comes from the GEF-Common project. Do you not get that one when you setup?

Evanosity: Nope, as far as I can tell the GEF-Common project isn't on Github.

Initium-Odp/src/com/universeprojects/miniup/CommonChecks.java
```diff
@@ -589,6 +589,19 @@ public static boolean checkLocationIsTopdownMode(CachedEntity location)
 		
 		return false;
 	}
+	
+	public static boolean checkItemHasAspect(CachedEntity item, String aspectName) {
+		
+		List<String> aspects = (List<String>) item.getProperty("_aspects");
+		if(aspects == null) return false;
+		
+		for(String aspect:aspects) {
```
Emperorlou, line 598: I think you can call .contains() here instead. It's easier

Initium-Odp/src/com/universeprojects/miniup/CommonChecks.java
```diff
@@ -599,7 +612,19 @@ public static boolean checkItemHasAspect(CachedEntity item, Class<? extends Init
...
 		return false;
 	}
-
+	
+	/**
+	 * Returns true if the given entity is embedded onto another entity.
+	 * @param entity
+	 * @return
+	 */
+	public static boolean checkEntityIsEmbedded(CachedEntity entity) {
```
Emperorlou, line 621: This should not be done this way. Remove this method

Initium-Odp/src/com/universeprojects/miniup/server/ODPDBAccess.java
```diff
@@ -678,6 +678,21 @@ public CachedEntity getEntity(Key key)
...
 		return pool.get(key);
 	}
+	
+	/**
+	 * Returns the parent of this entity.
+	 * Will return null if the given entity is not an embeddedentity.
+	 * @param entity
+	 * @return
+	 */
+	public CachedEntity getCachedEmbeddedEntityParent(CachedEntity entity) {
```
Emperorlou, line 688: This should not be done this way. Remove this method

Initium-Odp/src/com/universeprojects/miniup/server/ODPDBAccess.java
```diff
@@ -2651,6 +2666,25 @@ public int compare(CachedEntity o1, CachedEntity o2)
...
+	/**
+	 * Filters a list of entities by an aspect.
+	 * @param entities - the original list of entities
+	 * @param aspectName - the aspect to filter for
+	 * @return the filtered list of entities
+	 */
+	public List<CachedEntity> filterEntitiesByAspect(List<CachedEntity> entities, String aspectName){
+		List<CachedEntity> toReturn = new ArrayList<>();
+		
+		for(CachedEntity entity:entities) {
```
Emperorlou, line 2679: Do a null check here so it doesn't crash on null

Initium-Odp/src/com/universeprojects/miniup/server/aspects/AspectSlotted.java
```diff
@@ -1,40 +1,184 @@
...
 		super(object);
 	}
 
+	//Here, we need to properly set the link. use aspectpet as a base.
 	@Override
 	public List<ItemPopupEntry> getItemPopupEntries(CachedEntity currentCharacter)
 	{
-		return null;
+		if (CommonChecks.checkItemIsAccessible(object.getEntity(), currentCharacter) && hasOpenSlots()) {
+			ItemPopupEntry ipe = new ItemPopupEntry("Insert an item into the slots", "Insert a slottable item into this item.",
```
Emperorlou, line 40: Maybe "Insert an item into an available slot"?

Initium-Odp/src/com/universeprojects/miniup/server/aspects/AspectSlotted.java
```diff
@@ -1,40 +1,184 @@
...
+	}
+	
+	/**
+	 * Returns the number of open slots on this aspect.
+	 * @return
+	 */
+	public long getOpenCount() {
+		Long max = (Long)getProperty("maxCount");
+		
+		List<EmbeddedEntity> currentlySlotted = (List<EmbeddedEntity>)getProperty("slotItems"));
```
Emperorlou, line 83: If you wanted to get fancy you could add this get property call to it's own method on the aspect so you don't have to cast it all the time.

Initium-Odp/src/com/universeprojects/miniup/server/aspects/AspectSlotted.java
```diff
@@ -1,40 +1,184 @@
...
 		return count;
 	}
 	
-}
+	/**
+	 * Returns true if this aspect has any open slots.
+	 * @return
+	 */
+	public boolean hasOpenSlots() {
+		return getMaxCount() < getOpenCount();
```
Emperorlou, line 73: If open slots > 0 is probably what you're looking for here

Initium-Odp/src/com/universeprojects/miniup/server/aspects/AspectSlotted.java
```diff
@@ -1,40 +1,184 @@
...
+		@Override
+		public void run(Map<String, String> parameters) throws UserErrorMessage, UserRequestIncompleteException{
+			
+			//generate the base item from the given ID, then do basic checks.
+			Key baseKey = KeyFactory.createKey("Item", Long.parseLong(parameters.get("baseId")));
+			InitiumObject baseItem = db.getInitiumObject(baseKey);
+			
+			if(baseItem == null) throw new UserErrorMessage("Invalid Item ID.");
+			if(!CommonChecks.checkItemIsAccessible(baseItem.getEntity(), db.getCurrentCharacter()))
+				throw new UserErrorMessage("Nice try.");
```
Emperorlou, line 117: This can legitimately happen especially if you are using multiple tabs. Might be better to give a proper error message here

Initium-Odp/src/com/universeprojects/miniup/server/aspects/AspectSlotted.java
```diff
@@ -1,40 +1,184 @@
...
+			
+			if(slottedAspect == null) throw new UserErrorMessage("This item has no slots.");
+			
+			if(!slottedAspect.hasOpenSlots()) throw new UserErrorMessage("All slots on this item are full.");
+			
+			//generate the slottable item from the given ID, then do basic checks.
+			Key slottableKey = KeyFactory.createKey("Item", Long.parseLong(parameters.get("slottableId")));
+			InitiumObject slottableItem = db.getInitiumObject(slottableKey);
+			
+			if(slottableItem == null) throw new UserErrorMessage("Invalid Item ID");
```
Emperorlou, line 129: This can also be legitimate for the same reason especially with multiple tabs

Initium-Odp/src/com/universeprojects/miniup/server/aspects/AspectSlotted.java
```diff
@@ -1,40 +1,184 @@
...
+			if(!CommonChecks.checkItemHasAspect(baseItem.getEntity(), "Slottable"))
+				throw new UserErrorMessage("This item can't be inserted into a slot.");
+			
+			//we've made it this far, which means the base item and the slottable item are both completely valid.
+			//time to actually insert the item.
+			
+			//Write a method that copies a given entity to a new embeddedentity?
+			EmbeddedEntity newEmbedded = null;
+			
+			List<EmbeddedEntity> currentlySlotted = (List<EmbeddedEntity>)baseItem.getProperty("Slotted:slotItems");
```
Emperorlou, line 141: Probably better to use the InitiumObject object here and get the aspect object and manipulate that instead.

Initium-Odp/src/com/universeprojects/miniup/server/aspects/AspectSlotted.java
```diff
@@ -1,40 +1,184 @@
...
+			if(currentlySlotted == null) currentlySlotted = new ArrayList<>();
+			currentlySlotted.add(newEmbedded);
+			
+			List<String> aspectsToAdd = (List<String>) slottableItem.getProperty("Slottable:aspects");
```
Emperorlou, line 145: Everything from here down is highly suspect. We should have a chat about this stuff ^^

PR #971, opened 2020-10-08
Zombie fix

Evanosity:
Zombies should no longer be able to perform any action, with the exception of switching to a different character.


Initium-Odp/src/com/universeprojects/miniup/server/OperationBase.java
```diff
@@ -35,6 +38,10 @@ public OperationBase(HttpServletRequest request, ODPDBAccess db)
 	{
 		this.db = db;
 		this.request = request;
+		
+		if(CommonChecks.checkCharacterIsZombie(db.getCurrentCharacter()))
+			if(!(this instanceof CommandSwitchCharacter))
```
Emperorlou, line 43: Why would you do this if you already have the check in the command?

Evanosity: The command checks the target character, this checks the source character.

Emperorlou (approved):

PR #972, opened 2020-10-08
Charges visible

Evanosity:
We will need a schema change for this. If chargesVisible is set to true, they will display on the item. False by default.

### DIFF
--- a/Initium-Odp/src/com/universeprojects/miniup/server/controllers/ViewItemController.java
+++ b/Initium-Odp/src/com/universeprojects/miniup/server/controllers/ViewItemController.java
@@ -684,7 +684,7 @@ public class ViewItemController extends PageController {
 			itemMap.put("description", field.toString());
 		
 		//If we have explicitely allowed charges to be visible on this item, display them.
-		Boolean chargesVisible = (boolean) item.getProperty("chargesVisible");
+		Boolean chargesVisible = (Boolean) item.getProperty("chargesVisible");
 		if(chargesVisible == null) chargesVisible = false;
 		if(chargesVisible) {
 			field = item.getProperty("charges");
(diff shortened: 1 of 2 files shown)


Initium-Odp/src/com/universeprojects/miniup/server/controllers/ViewItemController.java
```diff
@@ -683,6 +683,16 @@ else if(ownerHtmlSpecified == false && GameUtils.enumEquals(scriptType, ODPDBAcc
 		if (field!=null && field.toString().trim().equals("")==false)
 			itemMap.put("description", field.toString());
 		
+		//If we have explicitely allowed charges to be visible on this item, display them.
+		Boolean chargesVisible = (boolean) item.getProperty("chargesVisible");
```
Emperorlou, line 687: this will cause  an error if charchesVisible is null. You have to cast it as a "Boolean", not "boolean". Boolean can be null, but boolean is a primitive and cannot be null.

Initium-Odp/src/com/universeprojects/miniup/server/controllers/ViewItemController.java
```diff
@@ -683,6 +683,16 @@ else if(ownerHtmlSpecified == false && GameUtils.enumEquals(scriptType, ODPDBAcc
 		if (field!=null && field.toString().trim().equals("")==false)
 			itemMap.put("description", field.toString());
 		
+		//If we have explicitely allowed charges to be visible on this item, display them.
+		Boolean chargesVisible = (boolean) item.getProperty("chargesVisible");
+		if(chargesVisible == null) chargesVisible = false;
+		if(chargesVisible) {
+			field = item.getProperty("charges");
```
Emperorlou, line 690: I find it odd that you're double indenting in all your code, can we use single indents? an indent should be done via the TAB button and it should be no more than 4 spaces wide per indent.

PR #973, opened 2020-10-08
Update all Commands/Long Operations to disallow zombies.

Evanosity:
-Disallow zombies from doing stuff, with a few exceptions.
-Clean up code related to calling the current character in a couple commands.

### DIFF
--- a/Initium-Odp/src/com/universeprojects/miniup/server/commands/CommandCharacterRespawn.java
+++ b/Initium-Odp/src/com/universeprojects/miniup/server/commands/CommandCharacterRespawn.java
@@ -29,6 +29,9 @@ public class CommandCharacterRespawn extends Command
 		
 		CachedEntity user = db.getCurrentUser();
 		CachedEntity deadCharacter = db.getCurrentCharacter();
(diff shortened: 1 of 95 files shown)


Initium-Odp/src/com/universeprojects/miniup/server/commands/CommandCharacterRespawn.java
```diff
@@ -29,6 +29,9 @@ public void run(Map<String, String> parameters) throws UserErrorMessage, UserReq
 		
 		CachedEntity user = db.getCurrentUser();
 		CachedEntity deadCharacter = db.getCurrentCharacter();
+
```
Emperorlou, line 32: I think you want to be able to do this one, no?

Evanosity: I don't think so, they get a new character with the same name when the character gets zombified. The zombie should just stay dead and do nothing.

Emperorlou (approved):

PR #974, opened 2020-10-12
Fixes for root location pathing

Evanosity:
-Characters will no longer see paths that they dont have access to from the global map.
-Standardize code for loading visible paths.

### DIFF
--- a/Initium-Odp/src/com/universeprojects/miniup/server/services/MainPageUpdateService.java
+++ b/Initium-Odp/src/com/universeprojects/miniup/server/services/MainPageUpdateService.java
@@ -722,8 +722,8 @@ public class MainPageUpdateService extends Service
 			}
 			else
 			{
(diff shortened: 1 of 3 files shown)


Initium-Odp/src/com/universeprojects/miniup/server/services/MainPageUpdateService.java
```diff
@@ -722,8 +722,8 @@ public String updateGlobalNavigationMap()
 			}
 			else
 			{
-				// Load the paths and destLocations for the root location
-				paths = db.getPathsByLocation_PermanentOnly(location.getKey());
+				// Load the paths and destLocations for the root location. Load only the paths that the user can see.
```
Emperorlou, line 725: What is the reasoning behind the change?

Emperorlou (approved):

PR #975, opened 2020-10-12
Improve Item Splitting Logic

Evanosity:
-Allow for the user to specify both a stack size and a number of stacks.
-Allow for the user to select multiple stacks of items to split (up to 10, a number I arbitrarily decided)

### DIFF
--- a/Initium-Odp/src/com/universeprojects/miniup/server/commands/CommandItemsStackSplit.java
+++ b/Initium-Odp/src/com/universeprojects/miniup/server/commands/CommandItemsStackSplit.java
@@ -13,6 +13,11 @@ import com.universeprojects.miniup.server.GameUtils;
 import com.universeprojects.miniup.server.ODPDBAccess;
 import com.universeprojects.miniup.server.commands.framework.UserErrorMessage;
 
+/**
+ * 
+ * @author Evan
+ *
+ */
 public class CommandItemsStackSplit extends CommandItemsBase {
 
 	public CommandItemsStackSplit(ODPDBAccess db, HttpServletRequest request, HttpServletResponse response) {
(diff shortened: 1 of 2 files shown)


Initium-Odp/src/com/universeprojects/miniup/server/commands/CommandItemsStackSplit.java
```diff
@@ -23,61 +28,93 @@ public CommandItemsStackSplit(ODPDBAccess db, HttpServletRequest request, HttpSe
...
+			
+			if (stackSize <= 0 || stacks <= 0) 
 				throw new UserErrorMessage("Please input a positive integer value.");
+			
+			if(batchItems.size() > 1 && stacks > 1) 
+				throw new UserErrorMessage("You can't split multiple items into multiple stacks.");
+			
+			if(stacks > 24) 
+				throw new UserErrorMessage("Too many stacks. Max is 24.");
+
```
SPFiredrake, line 55: Think about how many items you are creating if you have 10 items of 100 quantity each making 10 stacks of 10.

Initium-Odp/src/com/universeprojects/miniup/server/commands/CommandItemsStackSplit.java
```diff
@@ -23,61 +28,93 @@ public CommandItemsStackSplit(ODPDBAccess db, HttpServletRequest request, HttpSe
...
+					CachedEntity newStack = new CachedEntity(splitItem.getKind(), ds.getPreallocatedIdFor(splitItem.getKind()));
+					CachedDatastoreService.copyFieldValues(splitItem, newStack);
+					
+					newStack.setProperty("quantity", stackSize);
+					ds.put(newStack);
+				}
+				
+				splitItem.setProperty("quantity", newStackSize);
+				ds.put(splitItem);
+
```
SPFiredrake, line 98: What happens to the original item if it evenly divides into the stacks. 100 item, split into 10 stacks of 10. You'd end up with 11 items, one of which has 0 quantity.

SPFiredrake: Nevermind. Looks like you're preventing new stack size from being 0 by requiring that the stack can be created.

PR #983, opened 2020-10-28
Do a null check

Evanosity:


Initium-Odp/src/com/universeprojects/miniup/server/commands/CommandRefreshTile.java
```diff
@@ -25,6 +25,8 @@ public void run(Map<String, String> parameters) throws UserErrorMessage, UserReq
 		Integer tileX = parseInteger("selected2DTileX");
 		Integer tileY = parseInteger("selected2DTileY");
 		
+		if(tileY == null || tileX == null) return;
```
Emperorlou, line 28: I thought the "regenerateDBItemTileCache" call assumes the middle of the map if nothing is selected.

Emperorlou (approved):

PR #991, opened 2020-12-16
Update view_profile.jsp

Evanosity:


Initium-Odp/war/WEB-INF/odppages/view_profile.jsp
```diff
@@ -42,7 +42,7 @@
 	<a onclick='changeEmailAddress("${userEmail}")'>Change your account's email address</a>
 </p>
 <p>
-	<a href='resetpassword.jsp'>Reset password</a>
+	<a href='https://www.playinitium.com/resetpassword.jsp'>Reset password</a>
```
SPFiredrake, line 45: Do `/resetpassword.jsp` instead. We want to keep the URL root, in case domain ever changes.

Issue #964, opened 2020-09-10
enemy hp is hidden. 


CryBaby66:
![infinium 2](https://user-images.githubusercontent.com/71088664/92756799-25349d00-f35b-11ea-876d-3a8cdb594c16.PNG)

tried refreshing everything.

Emperorlou:
Wow that's pretty wack. Is this the only monster it happens to?

CryBaby66:
If its name sorter i get the hp bar where Afrit is located. If the name takes 2 lines i cant see a hp bar.

Emperorlou:
I think this is fixed now

PR #995, opened 2020-12-22
Partied Combat in Experimental Mode pushes the combatant down

Evanosity:


Evanosity:
Fixes https://github.com/Emperorlou/Initium-ODP/issues/964

PR #996, opened 2020-12-22
Nested container support

Evanosity:


Initium-Odp/src/com/universeprojects/miniup/server/services/ContainerService.java
```diff
@@ -50,9 +51,11 @@ else if (durability > 0) {
 	 * This is for stuff that allows access to a given container (location, item, or character). Returns false if access should not be allowed.
 	 * @param character
 	 * @param container
+	 * @param checkNested TODO
 	 * @return
+	 * @throws EntityNotFoundException 
 	 */
-	public boolean checkContainerAccessAllowed(CachedEntity character, CachedEntity container)
+	public boolean checkContainerAccessAllowed(CachedEntity character, CachedEntity container, boolean checkNested)
```
Emperorlou, line 58: I'm sorry, I don't think we want this. It's too expensive and the expense is the reason why it wasn't done to begin with.

Evanosity: How do you feel about only checking a single level of nested instead of 4?

Emperorlou:
I'm sorry, I don't think we want this. It's too expensive and the expense is the reason why it wasn't done to begin with.

I'm open for discussion though.

PR #997, opened 2020-12-22
Teaching Ideas

Evanosity:
Not yet finished.
- how do we mark that a character is in "learn" mode?
- how do we render an idea? is there already a method for it?
- how do we generate a new constructitemidea from a constructitemideadef?

### DIFF
--- a/Initium-Odp/src/com/universeprojects/miniup/server/longoperations/LongOperationTeachIdea.java
+++ b/Initium-Odp/src/com/universeprojects/miniup/server/longoperations/LongOperationTeachIdea.java
@@ -1,0 +1,119 @@
+package com.universeprojects.miniup.server.longoperations;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Map;
+
+import com.google.appengine.api.datastore.Key;
+import com.google.appengine.api.datastore.KeyFactory;
+import com.universeprojects.cacheddatastore.CachedEntity;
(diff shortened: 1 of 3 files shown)


Initium-Odp/src/com/universeprojects/miniup/server/longoperations/LongOperationTeachIdea.java
```diff
@@ -0,0 +1,118 @@
...
+		
+		QueryHelper query = new QueryHelper(db.getDB());
+		
+		for(CachedEntity ideaDef : ideaDefs) {
+			nextStudent:
+			for(CachedEntity student : students) {
+				//make sure the student is still valid. Ensure that they're still in learn mode, and that they're still in the same location.
+				
+				if(GameUtils.equals(student.getProperty("locationKey"), character.getProperty("locationKey")) == false) continue;
+				if(GameUtils.equals(student.getProperty("mode"), "LEARNING") == false) continue; //not sure if this is how we want to do this.
```
Emperorlou, line 93: I don't think the player enters a different mode when doing an invention do they? So we would want to do it just like invention.

Evanosity: Normal inventions are just a long operation, which locks the character inventing. But in this case, we are affecting other characters. How do we mark the characters that are being taught, to prevent them from doing other things while they are being taught?

PR #998, opened 2020-12-23
Add a boolean for armies to propagate forever

Evanosity:


Emperorlou:
I don't THINK i'd like to have this. You can just set a very high number, but at least it wouldn't propagate forever. This runs a risk of burning CPU/database cycles once everyone leaves and the monsters would just take over the whole world rendering the game unplayable.

PR #999, opened 2020-12-23
Knowledge and GameUtils.equals in script context

Evanosity:


Initium-Odp/src/com/universeprojects/miniup/server/scripting/wrappers/Character.java
```diff
@@ -399,4 +402,64 @@ public Group getGroup()
...
+	 * @param item
+	 * @param amount
+	 * @param percentageCap - what percent of the knowledge cap can you get from this increase?
+	 * @return
+	 */
+	public boolean increaseKnowledgeFor(Item item, int amount, int percentageCap) {
+		return getKS().increaseKnowledgeFor(item.getEntity(), amount, percentageCap);
+	}
+	
+	private ODPKnowledgeService getKS() {
```
Emperorlou, line 459: Lets not use contractions like this. DB is ok because people use DB in every day language. But KS is not immediately obvious what it's referring to.

Emperorlou (approved):

PR #1001, opened 2020-12-23
Pinning quests

Evanosity:
- Add a new command for pinning quests. When pinned, a quest will be forced to appear on the main page.
- Seperate the quest list into active and finished quests
- Don't show begin quest dialog on an item if it's inaccessible


SPFiredrake:
Don't know enough about the Quest system to make a determination here. @Emperorlou, can you please review?

Emperorlou (approved):

PR #1002, opened 2020-12-24
Quests in Script Context

Evanosity:


Initium-Odp/src/com/universeprojects/miniup/server/scripting/wrappers/Character.java
```diff
@@ -462,4 +460,72 @@ private ODPKnowledgeService getKnowledgeService() {
...
+		
+		return toReturn.toArray(new Quest[toReturn.size()]);
+	}
+	
+	private QuestService getQuestService() {
+		if(questService == null)
+			questService = db.getQuestService(null);
+		
+		return questService;
+	}
```
SPFiredrake, line 530: If possible, try to separate DA from the actual POCO objects. What if we add Quest support to items or locations? Having a separate location for the DA would require changing in only 1 place.

PR #1005, opened 2021-01-02
Depost all gold

Ashiix:
If the field in gold deposit is empty, automatically set it to the amount the player has to make it easier to deposit


Ashiix:
Ah I forgot to create a branch for the "Hide character widget while traveling" commit so I guess they're stuck together here.

PR #1010, opened 2021-01-21
Scripting helper methods

Evanosity:
- Allow for us to pull a container's contents from script context.
- helper methods in this are just to reduce code duplication inside scripts

### DIFF
--- a/Initium-Odp/src/com/universeprojects/miniup/server/scripting/wrappers/EntityWrapper.java
+++ b/Initium-Odp/src/com/universeprojects/miniup/server/scripting/wrappers/EntityWrapper.java
@@ -1,6 +1,5 @@
 package com.universeprojects.miniup.server.scripting.wrappers;
 
-import java.security.spec.DSAGenParameterSpec;
 import java.util.ArrayList;
 import java.util.Iterator;
 import java.util.List;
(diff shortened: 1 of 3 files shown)


Initium-Odp/src/com/universeprojects/miniup/server/scripting/wrappers/EntityWrapper.java
```diff
@@ -50,6 +47,18 @@ public boolean validMessageTarget() {
 		return false;
 	}
 	
+	public Key getDefKey() {
+		return (Key) getProperty("_definitionKey");
+	}
```
SPFiredrake, line 52: Rename to `getDefinitionKey`. Don't be lazy.

Initium-Odp/src/com/universeprojects/miniup/server/scripting/wrappers/Item.java
```diff
@@ -58,6 +58,23 @@ public boolean setContainer(EntityWrapper ent)
...
+		List<Item> toReturn = new ArrayList<>();
+		
+		for(CachedEntity ce : results) {
+			toReturn.add((Item) ScriptService.wrapEntity(ce, db));
+		}
+		
+		contents = toReturn.toArray(new Item[toReturn.size()]);
+		
+		return contents;
+	}
```
SPFiredrake, line 76: Make sure that the parent is not also an item, since we don't want to allow querying nested containers at all (to mimic the limitations of the ODP, don't want to create a workaround in scripting).

PR #1023, opened 2021-02-25
Fixes for #1021/#1022

Evanosity:
this is basically a PR authored by drake


Initium-Odp/war/odp/javascript/script.js
```diff
@@ -2473,13 +2473,14 @@ function switchCharacter(eventObject, characterId, direction)
 	direction = direction || 0;
 	doCommand(eventObject,"SwitchCharacter",{"characterId":characterId, "direction":direction},function()
 	{
+		var charParam = window.characterOverride != null ? ("?char=" + window.characterOverride) : "";
```
SPFiredrake, line 2476: Fix the null check here. Should be `window.characterOverride!=null && window.characterOverride.length>10`

PR #1027, opened 2021-04-30
Give embedded item popups some love

Evanosity:
* Allow for us to use embedded item popups inside scripting calls.
* Fix embedded item popups when looting instanced monsters.


Emperorlou:
What is happening in this PR?

PR #1028, opened 2021-04-30
Achievements in scripting

Evanosity:
* allow scripts to grant achievements.


Emperorlou:
You didn't handle if the achievements list on the user was null to begin with

Evanosity:
> You didn't handle if the achievements list on the user was null to begin with

I have now fixed this.

PR #1036, opened 2021-05-06
Automatically deposit all gold if no value is entered

Ashiix:
Automatically deposit all gold into a container if the value field is left blank.


Emperorlou (commented):
Cool, but can you just clean up the PR so it doesn't include all the whitespace changes?

PR #1040, opened 2021-05-10
Deposit all gold on character if no value is input

Ashiix:


Initium-Odp/src/com/universeprojects/miniup/server/commands/CommandDogeCoinsDepositToItem.java
```diff
@@ -58,18 +58,16 @@ public void runInsideTransaction(Map<String, String> parameters) throws UserErro
...
 
-		if (parameters.get("amount") == "")
-			depositAmount = characterCoins;
-		else
-			depositAmount = GameUtils.fromShorthandNumber(parameters.get("amount").replace(",", ""));
+		depositAmount = GameUtils.fromShorthandNumber(parameters.get("amount").replace(",", ""));
 
-		if(depositAmount == null)
-			throw new UserErrorMessage("Please type a valid gold amount.");
+		if(depositAmount == null || depositAmount == "")
```
Emperorlou, line 63: depositAmount is a Long so it could never equal ""

Also even if depositAmount was a string and if that string was "" it would still not equal "" because that's not how java does string comparisons. You'd want to do depositAmount.equals(""). But anyway, this is unecessary so you can remove this version of the check altogether.

Emperorlou (approved):

Initium-Odp/src/com/universeprojects/miniup/server/commands/CommandDogeCoinsDepositToItem.java
Ashiix: The original code was running the same comparison as an attempt to do what I did, wasn't sure why it was but didn't want to break the comparison in-case. Just rewrote it to make it actually work.

